PR #2, opened 2026-07-20
Add warning for models that may not fit in memory

lucasnewman:
It uses a fairly conservative estimate for now (80% of working memory needed) and will eventually need refinement to cover more quantization types, but this should be helpful for users to select models that are likely to work correctly.

Screenshot 2026-07-20 at 2 01 26 PM


Blaizzy (approved):
LGTM, thanks! 🚀

PR #6, opened 2026-07-21
Detect Hugging Face hub cache from HF_HOME/HF_HUB_CACHE

### LINKED ISSUE #5: Models in `HF_HOME` / `HF_HUB_CACHE` locations are not detected
## Problem

Nativ only scans the hardcoded `~/.cache/huggingface/hub` for local models.

Users who relocate their hub cache via the standard `huggingface_hub` environment variables (`HF_HOME` or `HF_HUB_CACHE`) don't see their models, and there is no setting in the UI to change the search path — the only workaround is hand-editing `~/Library/Application Support/Nativ/Settings.plist`.

## Expected

Resolve the default hub cache like the `huggingface_hub` Python library: `HF_HUB_CACHE` if set, else `$HF_HOME/hub`, else `~/.cache/huggingface/hub`.

Existing installs whose persisted setting is still the untouched legacy default should pick up the configured location automatically.

## Complication on macOS

GUI apps inherit launchd's environment, which excludes exports from shell startup files (`.zshrc`, …) — so reading `ProcessInfo` alone doesn't help the common case of a GUI-launched app with `HF_HOME` in `.zshrc`. 

A complete fix should also resolve the user's login-shell environment once at startup (the approach VS Code and JetBrains IDEs use).

> Note: this issue was generated by AI, curated by me.

rcanand:
Fixes #5 

## Summary

- Resolve the default model search path like the `huggingface_hub` Python library: `HF_HUB_CACHE` → `$HF_HOME/hub` → `~/.cache/huggingface/hub`.
- Existing installs that never customized the path have the legacy default persisted in Settings.plist; it is re-resolved against the environment on settings decode. Genuinely customized paths are always preserved.
- macOS GUI apps can't see shell rc-file exports, so when the process environment doesn't configure the cache, the app probes the user's login + interactive shell once at startup (async, 3s timeout — the same approach VS Code and JetBrains IDEs use) and adopts its `HF_HOME`/`HF_HUB_CACHE`. Parsing is
NUL-separated behind a marker segment, so noisy rc files (prompts, banners) can't corrupt values.
- Downloads and the embedded server already flow from this setting, so no other call sites change; views re-scan reactively when the path changes.

## Testing

- Adds a hostless `NativTests` XCTest target (the repo had none): 21 tests covering path-resolution precedence, legacy-default migration, `isConfigured`, env-output parsing (marker, junk segments, values with `=`/newlines), and process execution (output capture, timeout, missing executable). `xcodebuild
-scheme Nativ -configuration Debug test` — all pass.
- Manually verified on a machine with `HF_HOME` only in `~/.zshrc`, launching via `open` (i.e. launchd environment):
  - existing-user upgrade (persisted legacy default): Models view flips from `~/.cache/huggingface/hub` contents to the `HF_HOME` cache within ~1–3s, and the migrated path persists across relaunches;
  - fresh install (no plist): same result;
  - terminal launch with `HF_HOME` exported: detected instantly, probe skipped.
- Probe failures (missing shell, timeout, rc errors) degrade gracefully to current behavior.

## Notes for reviewers
- Run `xcodegen generate` after checkout to get the test target; the tracked `Nativ.xcodeproj` is intentionally unchanged to avoid xcodegen-version churn (CI regenerates it).

Note: this fix was generated by AI, verified by me.

### DIFF
--- a/Sources/Nativ/Features/Models/HuggingFaceCache.swift
+++ b/Sources/Nativ/Features/Models/HuggingFaceCache.swift
@@ -1,0 +1,53 @@
+import Foundation
+
+/// Locates the Hugging Face hub cache using the same environment variables
+/// as the `huggingface_hub` Python library: `HF_HUB_CACHE` takes precedence,
+/// then `HF_HOME` (with `/hub` appended), then a per-user default.
+enum HuggingFaceCache {
+    /// Cache location used when neither environment variable is set.
+    static let fallbackHubPath = "~/.cache/huggingface/hub"
+
+    /// The default hub cache path for the given environment.
+    static func defaultHubPath(
+        environment: [String: String] = ProcessInfo.processInfo.environment
+    ) -> String {
+        if let cachePath = nonEmpty(environment["HF_HUB_CACHE"]) {
+            return cachePath
+        }
+        if let homePath = nonEmpty(environment["HF_HOME"]) {
+            return (homePath as NSString).appendingPathComponent("hub")
+        }
+        return fallbackHubPath
+    }
+
+    /// The effective model search path given a persisted setting.
+    ///
+    /// Installations that never customized the search path still have the
+    /// legacy hardcoded default persisted; re-resolve those against the
+    /// environment so `HF_HOME` and `HF_HUB_CACHE` are honored.
+    static func resolvedSearchPath(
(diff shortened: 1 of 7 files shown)


lucasnewman (approved):
Thanks!

PR #8, opened 2026-07-21
fix: don't report multi-shard models installed until fully downloaded

Hayden1629:
## Problem

Clicking **Install** on a model, then trying to run it before the download finishes, showed the model as **Installed** but running it failed with **"can't connect to server."** And once a partial download was correctly hidden, there was no way to cancel or clean it up from the UI.

## Root cause

Installed-state is derived from a filesystem scan (`installedModelIDs` → `LocalModelDiscovery.scan`), which calls `isLikelyMLXModelSnapshot` to decide whether a cached repo counts as installed. That check short-circuited to `true` the moment `model.safetensors.index.json` existed:

```swift
let indexURL = snapshotURL.appendingPathComponent("model.safetensors.index.json")
if fileManager.fileExists(atPath: indexURL.path) {
    return true
}
```

The index is a tiny JSON file that `huggingface_hub`'s `snapshot_download` fetches **before** the weight shards it points to. So for any multi-shard model, the snapshot looked "installed" after just the config + index landed, while the actual weights were still downloading. Selecting it started `mlx-vlm-server` against missing weights; the server failed to load the model and never bound its port, surfacing to the client as "can't connect to server."

## Changes

**1. Fix the false "installed" state** (`fix:` commit)

Validate the index against the files on disk: parse `weight_map`, collect the referenced shard filenames, and only treat the snapshot as installed once **every** shard is present. Single-file and non-safetensors layouts are unaffected — the new helper returns `nil` and falls through to the existing presence logic.

**2. Let partial downloads be managed from the Installed page** (`feat:` commit)

Hiding partials fixed the bug but left them unmanageable, so this surfaces them where the user already looks:

- `LocalModelDiscovery.scanAll` classifies each cached repo as **installed**, **incomplete** (sharded index present but shards missing, or a leftover `*.incomplete` blob), or skipped. `scan()` still returns `[LocalModel]`, so existing callers (`AppDelegate`, `DashboardViewModel`) are unchanged.
- `LocalModelLibrary` publishes `incompleteModels` and gains a `repoID`-based `delete` (the model-based `delete` delegates to it).
- The Installed page renders a `PendingModelRow` for **active downloads** (live progress + pause/resume, from the download manager) and **incomplete leftovers** (from the cache scan), de-duplicated against installed models. The delete/cancel button is always enabled — cancel routes to `removeDownload` for active transfers and deletes the partial files for leftovers.

## Testing

- `make xcode-build` — succeeds.
- `make xcode-smoke` — passes.
- Manual: reproduced a multi-shard partial in the HF cache — it no longer shows as installed, appears on the Installed page as "Incomplete download," and its delete button removes it. A live download shows on the Installed page with progress and cancels cleanly.
- Verified the pure logic against synthetic snapshot dirs:

  | Case | Result |
  |---|---|
  | index only, no shards | not installed / incomplete |
  | index + shard 1 of 2 | not installed / incomplete |
  | index + all shards | installed |
  | single-file, `.incomplete` blob present | incomplete |
  | no index (single-file) | falls through to existing logic |

Note: the repo has no Swift test target, so I verified the logic standalone rather than adding a test. Happy to add a test target if you'd like one.

### DIFF
--- a/Sources/Nativ/Features/Models/LocalModelDiscovery.swift
+++ b/Sources/Nativ/Features/Models/LocalModelDiscovery.swift
@@ -133,6 +133,27 @@ struct LocalModel: Identifiable, Equatable, Sendable {
     }
 }
 
+/// A model repository present in the cache that has not finished downloading —
+/// either currently in progress or a leftover partial from an interrupted run.
+struct IncompleteLocalModel: Identifiable, Equatable, Sendable {
+    var id: String { repoID }
+
+    let repoID: String
+    let provider: LocalModelProvider?
+    let downloadedBytes: Int64?
+    let snapshotURL: URL?
+    let modifiedAt: Date?
+}
+
+/// The outcome of a cache scan, split into fully installed models and partial
+/// downloads so the UI can surface and manage each independently.
+struct LocalModelScanResult: Equatable, Sendable {
+    let installed: [LocalModel]
+    let incomplete: [IncompleteLocalModel]
+
+    static let empty = LocalModelScanResult(installed: [], incomplete: [])
+}
+
 struct LocalModelMemoryEstimate: Equatable, Sendable {
     static let headroomFraction = 0.20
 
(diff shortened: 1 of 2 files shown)


Lazarus-931:
hi @Hayden1629, there seems to be merge conflict, resolve and will take a look!

lucasnewman:
Covering this in https://github.com/Blaizzy/nativ/pull/225

PR #18, opened 2026-07-21
fix: keep model downloads running when leaving the Models page

### LINKED ISSUE #17: Downloads stop when you leave the Models page
Start a model download on the Models page, then switch to any other tab (Chat, Dashboard, …) — the download is cancelled and progress is lost.

Cause: `ModelsView` owns `HuggingFaceDownloadManager` as a view-scoped `@StateObject` and calls `downloadManager.cancelDownload()` in `onDisappear` (`ModelsView.swift:62`), so navigating away tears down the in-flight download. The manager needs to live at app scope for downloads to survive navigation.

Lazarus-931:
Fixes #17. The download manager was view-scoped and cancelled in `onDisappear`, so leaving the Models page killed active downloads. Share one app-scoped instance so downloads survive navigation.


lucasnewman (approved):
LGTM

PR #19, opened 2026-07-21
feat: allow choosing the local server port

### LINKED ISSUE #9: Feature Request: Support custom local port for server endpoints (instead of hardcoded 8080)
I really enjoy using nativ, but I run into port conflicts because it uses 8080 as the default (and seemingly fixed) local port.

Port 8080 is one of the most commonly used ports by developers — it's the default for many frameworks (Spring Boot, Tomcat, Node.js/Express dev servers, etc.), so it's often already occupied by other local services on my machine. Every time I want to start your app, I have to stop another service first, which is quite inconvenient.

Please add support for a custom local port so users can easily configure it. 
This would make the nativ much more developer-friendly and avoid unnecessary port conflicts.

Thanks for considering this feature! Happy to provide more details or test any implementation if helpful.

Best regards,
Barat.

Lazarus-931:
This closes #9 to allow devs to add a server of their own for model endpoints, which before used to default to 8080, which is restrictive with multiple local servers.

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatConfigurationView.swift
+++ b/Sources/Nativ/Features/Chat/ChatConfigurationView.swift
@@ -29,6 +29,7 @@ struct ModelConfigurationLayout<Content: View>: View {
                 ModelConfigurationView(
                     settings: $model.settings,
                     settingsRequireRestart: model.settingsRequireRestart,
+                    activeServerPort: model.activeServerPort,
                     onReset: model.resetSettings
                 )
                 .frame(width: 320)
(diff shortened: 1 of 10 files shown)


Sources/Nativ/Features/Chat/ChatConfigurationView.swift
```diff
@@ -125,6 +129,41 @@ struct ModelConfigurationView: View {
         .padding(.vertical, 16)
     }
 
+    private var serverSection: some View {
```
lucasnewman: Would this be a better fit in the Developer section? It's more of a global config than a per-chat thing.

lucasnewman:
@Lazarus-931 You need to run `xcodegen` locally to pick up the new file in the project or it will break the pre-built Xcode project. Other than that, LGTM

lucasnewman (approved):

Lazarus-931:
@lucasnewman moved the port config to Developer page instead of settings

PR #20, opened 2026-07-21
Add dev build verification workflow for every branch and PR.

Lazarus-931:
This setups up gh actions debug builds for every branch & pr


lucasnewman:
@Lazarus-931 This would result in an unsigned build, which isn't usable without re-signing it (or jumping through lots of OS hoops). Maybe we just do the build step and skip the output artifact?

Lazarus-931:
@lucasnewman, yeah didn't realize this, will just make it build verification

lucasnewman (approved):
Awesome, thanks!

PR #21, opened 2026-07-21
Detect MLX models installed by LM Studio.

Lazarus-931:
Scans ~/.lmstudio/models (and the legacy ~/.cache/lm-studio/models) for MLX-format model directories, lists them on the Models page with an LM Studio badge, and makes them selectable in chat by loading the model from its directory path. External models are never deleted by Nativ.

### DIFF
--- a/Sources/Nativ/AppDelegate.swift
+++ b/Sources/Nativ/AppDelegate.swift
@@ -852,9 +852,14 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuDelegate {
         return NumberFormatter.localizedString(from: NSNumber(value: value), number: .decimal)
     }
 
+    private var modelScanKey: String {
+        let settings = model.settings.normalized()
+        return ([settings.expandedModelSearchPath] + settings.additionalModelSearchPaths)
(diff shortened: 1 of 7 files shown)


lucasnewman (approved):
LGTM

Blaizzy (changes requested):
Why don't we allow users to pass custom paths that we save to DB instead of hard coding new paths to services we don't own or are sure will remain in the same directory. 

This approach is more robust and will allow users to even pass external SSDs as model sources. 

And we keep the HF on by default. 

Resulting in HF cache + custom paths.

Blaizzy:
Perfect, could you share a screenshot of how it looks like?

Lazarus-931:
tried but don't have enough mem to download full xcode for build, if we merge #20, gh actions can build and i can download to test any prs we make

Blaizzy (approved):
LGTM, thanks!

Blaizzy:
No worries, great work! 

merging it 🚀

Blaizzy:
No worries, great work! 

merging it 🚀

Blaizzy:
Please fix conflicts and we can merge

PR #37, opened 2026-07-21
Fix software update

lucasnewman:
The URL was pointing at the staging repo and the website deployment was overriding the appcast, so the current scheme doesn't work. This changes the release deployment to include the appcast as a release artifact and then the app can pull the latest release from there.


Blaizzy (approved):
LGTM, thanks!

PR #40, opened 2026-07-22
Fix Discover pagination showing inconsistent page sizes

Lazarus-931:
Model search fetched raw 50-item API pages, then filtered them client-side, so each displayed page showed however many survived the filter (e.g. 3, then 10). Back paging with a flat filtered buffer sliced into fixed 24-item pages, fetching more raw pages as needed to fill each one.


lucasnewman (approved):
LGTM

PR #42, opened 2026-07-22
Add Copy Conversation and Export as Text to chat context menu

Lazarus-931:
Right-clicking a chat in the sidebar now offers copying the whole conversation to the clipboard or exporting it as a plain-text file, alongside Open and Delete. Text is rendered speaker-by-speaker from the stored session.


lucasnewman (approved):
Nice!

Issue #46, opened 2026-07-22
Image-gen fit estimate is weights-only and misleads for diffusion


Lazarus-931:
Part of #44.

**What**
`LocalModelMemoryEstimate` (`HuggingFaceModel.memoryEstimate` / `LocalModel.memoryEstimate`) sizes a model by resident weights only, and `isUsable` compares that against the memory budget. For image-generation pipelines that is wrong in the dangerous direction: the activation peak (attention + VAE decode) dominates, is roughly dtype-independent, and grows with output resolution — so a 4-bit diffusion repo can pass "Likely fits in memory" and then OOM at generation. Multi-component repos (`model_index.json`: text encoder + transformer + VAE) also lack an aggregate weight size from the hub `safetensors` field, so the pre-download number is off.

The mlx-vlm side already stages components (load → use → evict) and tiles VAE decode; the gap is that the Swift fit estimate doesn't reflect any of that.

**Approach**
- Add a peak-activation reserve to the estimate for `.imageGeneration` models (per-family constant to start; optionally scaled by resolution), folded into `isUsable`, keeping displayed weights honest.
- Multi-component pre-download size: walk the repo's component weight files instead of the single top-level `safetensors.parameters`.

**How to verify**
- A FLUX-class repo that OOMs at 1024×1024 on a 16 GB Mac shows "May not fit in memory" at that resolution; not over-warned at 512×512.
- A `model_index.json` repo reports size equal to the sum of its component weight files, not just the transformer.
- LLM/VLM fit labels unchanged (spot-check two text models).

Issue #44, opened 2026-07-22
Image-generation model support: memory-architecture suggestions & gotchas from shipping MLX diffusion pipelines


xocialize:
Hi — first, thanks for nativ (and for mlx-vlm/mlx-audio, which several of our Swift-MLX
ports build on). We've spent the last year shipping MLX diffusion/video pipelines on Apple
silicon (FLUX-class image models, Wan-family video, multi-component 3D texture pipelines),
and since the README lists dedicated image-generation model support as coming soon — and the
Swift side of the app (ImageGeneration feature, `/v1/images/*` clients, diffusion capability
detection in `LocalModelDiscovery`) already looks ready for it — we wanted to share the
things that bit us hardest on the serving/memory side, in case any of it saves you time.
Offering these as suggestions, not asks; happy to elaborate or contribute on any of them.

## The big one: weights-based memory estimates break for diffusion

Nativ's "likely fits in memory" estimate (params × bytes/param + headroom) works well for
LLMs, but for diffusion pipelines **activation memory dominates and is invisible to that
estimate**. Two properties we've measured repeatedly:

- **Activation peak is roughly dtype-independent.** Quantizing the transformer shrinks
  resident weights a lot but barely moves the transient peak (attention/scratch buffers are
  computed in fp16/fp32 regardless). On a large video DiT we measured ~12–15 GB of
  activation whether the weights were bf16, int8, or int4. A weights-only estimate will
  green-light an int4 pipeline that OOMs at generation time.
- **The peak is resolution-dependent**, driven mostly by VAE decode. A model that's fine at
  512×512 can fall over at 1536×1536 with identical weights. If the "may not fit" pill stays
  weights-only for image models, it will be wrong in the dangerous direction.

Suggestion: model image-gen memory as **resident weights + peak activation (a function of
output resolution)**, even if the activation term starts as a per-family constant. And since
components have disjoint lifetimes, the honest total is `max(phase)` not `sum(components)` —
see next point.

## Multi-component pipelines: per-stage load → use → evict

Image repos are `model_index.json` pipelines (text encoder + transformer + VAE), and the
components never need to be resident simultaneously: encode the prompt, **evict the text
encoder**, denoise, **evict the transformer**, decode. On our ports this is the difference
between "needs a 64 GB Mac" and "runs on 16–32 GB" — e.g. one bf16 video pipeline went from
an ~84 GB working set to ~40 GB from staging alone, no quality change. If mlx-vlm's image
[...]

PR #47, opened 2026-07-22
Reserve peak activation in the fit estimate for image models

Lazarus-931:
Diffusion pipelines are dominated by activation memory (attention + VAE decode), which is roughly dtype-independent and invisible to a weights-only estimate, so an int4 image pipeline could be labelled "fits" and then OOM at generation. LocalModelMemoryEstimate now carries an activation reserve (a coarse per-family constant), applied to image-generation models in both the hub and local producers, so the fit check reflects the working set rather than resident weights alone. Non-image models are unchanged.

addresses sub-issue #46, originally raised in #44 by @xocialize

cc: @lucasnewman


lucasnewman (approved):
LGTM

Blaizzy:
LGTM, 

Just wonder how much of it can be solved with chunking.

In MLX-Video the overheard to generate videos with the chunking algo is ~5-15GB and can be modulated by the chunk-size.

PR #48, opened 2026-07-22
Coalesce streamed chat updates to keep the UI responsive

Lazarus-931:
When each streamed token is applied to the messages component on the main actor, the constant updates saturate the run loop, so live UI like the transcript, the thinking bubble, and the "Working" animation stops repainting until some input event nudges it (issue #11). Buffering the content, reasoning, and metric deltas and flushing them to the view at ~30 Hz moves those writes off the per-token path, so the main thread is no longer starved during generation. A trailing flush guarantees the last partial chunk still lands, and a final flush on completion or cancellation applies whatever remains in the buffer. The result is a transcript and thinking bubble that update smoothly the whole way through a response instead of freezing mid generation.


lucasnewman (approved):
LGTM

PR #49, opened 2026-07-22
Allow custom HF tokens, pull HF_TOKEN in by default

lucasnewman:
This allows accessing gated models from HF and using a custom token in the Developer view.


Blaizzy (approved):
LGTM, thanks!

PR #50, opened 2026-07-22
Show recommended models during onboarding

Blaizzy:
## Summary
- show compatible Hugging Face MLX recommendations when onboarding finds no installed language models
- let users download, cancel, monitor, and select a first model without leaving onboarding
- prioritize models likely to fit the Mac while retaining load-on-demand
- fix the model-library task ownership and pagination compile errors exposed by the CI build

## Root cause
The onboarding picker only scanned the local cache. On a fresh installation, that left users with no model choices and no path to acquire one.

The CI failure came from `HuggingFaceHub.swift`: weak `self` was used as non-optional, the next-page URL was not unwrapped, and a newly required token argument was omitted.

## Validation
- `make xcode-build` passed
- test suite passed: 27 tests, 0 failures


lucasnewman (approved):
LGTM

PR #53, opened 2026-07-23
Fix macOS app icon rendering

Blaizzy:
## What changed

- add a native Icon Composer app-icon package targeted specifically at macOS
- regenerate the legacy app-icon assets from an opaque canvas so macOS owns the final mask
- remove the runtime `applicationIconImage` override that bypassed the system icon stack
- register the new icon resource and synchronize the existing Swift source references in the Xcode project

## Why

The app was supplying a pre-shaped legacy icon and replacing the bundle icon at runtime. On current macOS versions, that path is treated as compatibility artwork, producing an oversized black square inside a gray plate in the Dock and during launch/quit animations.

The app now lets macOS render the native icon composition and apply its own mask, sizing, and lifecycle effects.

## Impact

Nativ now displays a correctly sized black rounded icon without the gray border in the Dock, Applications, and app launch/quit transitions.

## Validation

- `xcodebuild -project Nativ.xcodeproj -scheme Nativ -configuration Debug -derivedDataPath build/IconVerify CODE_SIGNING_ALLOWED=NO build`
- launched the resulting Debug app
- visually verified the live Dock icon after clearing stale Launch Services/icon caches


Lazarus-931:
lgtm!

Lazarus-931 (approved):

PR #67, opened 2026-07-23
Add Images tab for image generation / editing

lucasnewman:
As part of this, I had to be much more rigorous about model kind detection across language/image/speech models. We now support having models across different types preloaded using the underlying feature in mlx-vlm. Image gen/edit as a tool in Chat for use with LMs that support tool calling will be a follow up feature.

Screenshot 2026-07-23 at 2 57 07 PM
Screenshot 2026-07-23 at 2 57 00 PM


Lazarus-931:
this is super cool @lucasnewman, just a question, seems like it preloads a LM, image and voice model without a RAM check, could this cause a OOM?

lucasnewman:
> this is super cool @lucasnewman, just a question, seems like it preloads a LM, image and voice model without a RAM check, could this cause a OOM?

Yes, good call. Let me add a specific warning for that.

Lazarus-931 (commented):
lgtm!

PR #68, opened 2026-07-23
Fix header styling to be consistent across all views

lucasnewman:
I also ran `xcodegen` to pick up the files that were added in the Xcode project.


Blaizzy (approved):
LGTM, thanks!

PR #79, opened 2026-07-24
Chat: centered logo in the empty transcript and an accent hue around the composer in chat page

Lazarus-931:
looks like this:
Screenshot 2026-07-23 at 8 45 11 PM

and in white
Screenshot 2026-07-23 at 8 45 31 PM


lucasnewman:
@Lazarus-931 Can we only use the accent color in dark mode? It doesn't make as much sense in light mode, I think it looks a bit worse that way.

lucasnewman (approved):

Blaizzy:
LGTM! One small change on my end — let's revert the accent color around the chat box to keep the design sleeker.

I added color hierarchy that naturally surfaces the chatbox in #111

PR #88, opened 2026-07-24
Add detailed diagnostics to issue reports

Lazarus-931:
Makes the in-app issue reporter produce reports detailed enough to triage without a round-trip.

- **Parsed crash report** — when a recent `Nativ*.ips` is present, the report now includes the exception type + signal, the termination reason (the human-readable cause, e.g. "Too many open files"), and the crashed thread's top frames, and copies the full crash report to the clipboard to paste in. Previously it only listed a short filename excerpt.
- **Richer runtime metrics** — adds the runtime snapshot and all-time stats alongside the existing session and latest-request metrics.
- **Server configuration** — includes the exact launch flags the server was started with, so config-dependent issues are reproducible.

No secrets are exposed: the API key and Hugging Face token live in the environment, not the launch arguments, and the whole body is still run through home-directory redaction.

This brings automated reports up to the detail of a hand-pasted crash log — like the recent file-descriptor-exhaustion report — so maintainers get the cause and backtrace up front.

looks like:

Screenshot 2026-07-24 at 12 16 17 PM

### DIFF
--- a/Sources/Nativ/ControlPanelView.swift
+++ b/Sources/Nativ/ControlPanelView.swift
@@ -377,9 +377,11 @@ struct ControlPanelView: View {
             sections: IssueDiagnostics.collect(category: category, model: model, runtime: runtime),
             serverOutput: IssueDiagnostics.serverOutputTail(model: model)
         )
-        if body.count > IssueReportBuilder.urlBodyCharacterBudget {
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
LGTM

Issue #85, opened 2026-07-24
[BUG] "ERROR: Application startup failed. Exiting."


stephanj:
Downloaded successfully Qwen3.7-27B-NVFP4 model but server doesn't start. 

<img width="532" height="252" alt="Image" src="https://github.com/user-attachments/assets/899c7782-a52c-4a96-bdcb-9dd375d1a96f" />

<img width="767" height="661" alt="Image" src="https://github.com/user-attachments/assets/2cf4f826-f2c2-48cb-b602-8680db61ebc9" />

Als your "Check for updates..." doesn't work.

Lazarus-931:
hi @stephanj, thanks for the issue. yeah, so seems to be a upstream mlx-vlm issue not being able to reconstruct that NVFP4 layout u see there(i.e weight_packed, weight_scale, input_global_scale, weight_global_scale), so without tensors loading, server crashes. Will address that there, feel free to open an issue there as well!

As for the update, we don't have any released so not to worry. Our release doesn't have a appcast.xml, which the feed URL directs to, leading to the error you see. We'll probably set it up in so releases do that moving forth!

Lazarus-931:
#92 merged, should be good to go for upcoming release @stephanj!

stephanj:
Build your new (merged) build but still the same error.

```
... 
language_model.model.layers.9.mlp.gate_proj.weight_scale,
language_model.model.layers.9.mlp.up_proj.input_global_scale,
language_model.model.layers.9.mlp.up_proj.weight_global_scale,
language_model.model.layers.9.mlp.up_proj.weight_packed,
language_model.model.layers.9.mlp.up_proj.weight_scale.

ERROR:    Application startup failed. Exiting.

mlx-vlm-server stopped with status 3
```

Lazarus-931:
hi @stephanj, I meant the update , [mlx-vlm#1706](https://github.com/Blaizzy/mlx-vlm/issues/1706) stlll open, will have a fix for model issue mentioned!

Lazarus-931:
merged, should see it in release!

lucasnewman:
I verified this works in the current release.

stephanj:
Indeed, thanks for fixing this!

PR #92, opened 2026-07-24
Allow attaching the signed appcast to an existing release via workfllow

Lazarus-931:
The release workflow only ran on release publish, so a release created outside it (e.g. the first release) has no appcast.xml, and the Sparkle feed at releases/latest/download/appcast.xml returns 404.

Add a workflow_dispatch trigger that takes a tag input and reuses the same build/notarize/appcast/upload path against that existing tag, resolving the tag, upload URL, and release notes from the GitHub API instead of the release event payload, and replacing any same-named asset so re-runs are idempotent.

for issue #85

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,6 +3,11 @@ name: Publish macOS Release
 on:
   release:
     types: [published]
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
🙏

PR #93, opened 2026-07-24
Add support for image generation and editing in chat

lucasnewman:
This plumbs through image generation and edit tools to the server when an image model is loaded, so we can use image generation in chat if the model supports tool calling:

Screenshot 2026-07-24 at 9 31 09 AM
Screenshot 2026-07-24 at 9 31 14 AM


Lazarus-931:
this is nice @lucasnewman , don't know if it' already inside, but like if you hover over the img, a download on bottom right of the img would be nice, with like a light hue to it

Lazarus-931 (approved):

Lazarus-931:
besides that, lgtm!

Issue #84, opened 2026-07-24
Crash due to file descriptor exhaustion


nevali:
i installed Nativ and did some experiments with one the Qwen models, and left it running but idle overnight

i just received a crash notification from macOS reporting FD exhaustion, so there's apparently a file descriptor leak somewhere:

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------
Process:             Nativ [2917]
Path:                /Applications/Nativ.app/Contents/MacOS/Nativ
Identifier:          io.github.blaizzy.nativ
Version:             0.0.1 (202607201707)
Code Type:           ARM-64 (Native)
Role:                Background
Parent Process:      launchd [1]
Coalition:           io.github.blaizzy.nativ [2495]
User ID:             502

Date/Time:           2026-07-24 11:14:45.7693 +0100
Launch Time:         2026-07-23 17:53:53.5368 +0100
Hardware Model:      Mac14,12
OS Version:          macOS 26.5.2 (25F84)
Release Type:        User

Crash Reporter Key:  1174DCD6-25D3-5F4E-10F1-37274515A692
Incident Identifier: 2A017F60-22A0-4323-AA1C-5CECEED1C9FB

Time Awake Since Boot: 63000 seconds

System Integrity Protection: enabled

Triggered by Thread: 4, Dispatch Queue: CA::CG::Queue

Exception Type:    EXC_CRASH (SIGABRT)
Exception Codes:   0x0000000000000000, 0x0000000000000000

Termination Reason:  Namespace COREANIMATION, Code 3, 
metal library file /System/Library/Frameworks/QuartzCore.framework/Resources/default.metallib failed to open with error: Too many open files


Thread 0::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x188853a74 kevent_id + 8
1   libdispatch.dylib             	       0x1886fda64 _dispatch_kq_poll + 220
2   libdispatch.dylib             	       0x1886fe474 _dispatch_event_loop_wait_for_ownership + 460
3   libdispatch.dylib             	       0x1886ea6ac __DISPATCH_WAIT_FOR_QUEUE__ + 432
4   libdispatch.dylib             	       0x1886ea1f8 _dispatch_sync_f_slow + 140
5   QuartzCore                    	       0x1929ba764 CABackingStoreGetFrontTexture(CABackingStore*, CGColorSpace*) + 128
[...]

PR #95, opened 2026-07-24
Invalidate client URLSessions on deinit to fix file-descriptor leak

Lazarus-931:
The chat, image, and metrics clients each create an ephemeral URLSession but never invalidate it, so its underlying sockets/file descriptors are retained until process exit. The metrics client is recreated on server start and polled while the app idles, so descriptors accumulate over a long-running session until the process hits its open-file limit, causing  a CoreAnimation "Too many open files" crash seen in #84.

This includes deinit so descriptors are released when the client is deallocated.


lucasnewman (approved):
Good catch!

PR #97, opened 2026-07-25
Add configurable server host

Blaizzy:
## Summary

- add a persisted server host setting beside the existing configurable port
- pass the selected host to the server and use it for clients, copied endpoints, active integration profiles, availability checks, and lifecycle smoke tests
- support hostname, IPv4, and IPv6 URL generation while retaining `127.0.0.1` as the backward-compatible default
- show a live countdown and restart a running server three seconds after the latest host or port edit without retriggering on page navigation
- distinguish an address already in use from an invalid host or port in endpoint warnings
- preserve the replacement process and active endpoint when the previous server’s termination callback arrives late
- add color accents to the Developer panel headers for clearer visual separation
- detect existing Hugging Face CLI login credentials without exposing the token, including custom `HF_TOKEN_PATH`, `HF_HOME`, and XDG locations
- show the active Hugging Face credential source with a masked suffix and token length
- organize Hugging Face authentication into status, active credential, and custom override sections with replace and logout controls
- update integration coverage and documentation for dynamic hosts

## Why

The server port was configurable, but the bind host and several derived URLs were still hard-coded to `127.0.0.1`. This prevented users from binding Nativ to another local interface or wildcard address and caused generated integration endpoints to ignore the intended host.

## Impact

Users can configure both host and port from the Developer page. Changing either automatically restarts a running server after a three-second debounce, while an offline server remains offline. Existing settings continue to use the loopback default.

## Validation

- `xcodebuild -project Nativ.xcodeproj -scheme Nativ -configuration Debug -derivedDataPath build/XcodeDerivedData CODE_SIGNING_ALLOWED=NO -skipPackageUpdates test`
- 73 tests passed
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Use Nativ as a private chat app, a model manager, a performance dashboard, or an
 | **System monitor** | Inspect live per-core CPU load, GPU utilization, unified memory and swap pressure, disk throughput, capacity, and SMART health. |
 | **Local APIs** | OpenAI-compatible chat, Responses, image, audio, and model endpoints, plus Anthropic Messages endpoints. |
 | **Coding-tool integrations** | Configure and launch Codex, Claude Code, Pi, Hermes, and OpenCode against models served by Nativ. |
-| **Developer workspace** | Set the server port, add a Hugging Face token for gated models, inspect runtime details, copy endpoint URLs, search and filter live server logs, and monitor server health. |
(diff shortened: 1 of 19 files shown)


Lazarus-931:
@Blaizzy do you have a screenshot of the UI change to see?

Blaizzy:
Screenshot 2026-07-25 at 2 37 24 PM
Here you go @Lazarus-931!

Please don't merge after approval this time, I'll do it after some final changes

Lazarus-931:
understood, this looks nice @Blaizzy!

Issue #71, opened 2026-07-23
Downloading models not pinned


backerman:
If the user leaves the discover models screen, the fact that there is a model downloading is not present in the interface outside of that model's entry in the list (which may not be present if the search was cleared).

Lazarus-931:
hi @backerman, I think a small UI on the top across pages to show model downloading progress would be useful, feel free to open pr, we'll get to it as well!

Lazarus-931:
closing this since #99 is merged!

PR #99, opened 2026-07-25
Surface active model downloads outside the Discover list

Lazarus-931:
Today an in-progress model download is only visible inline in the **Discover** list. Switching tabs rebuilds `ModelsView` (resetting its section/filters) while the download keeps running in the shared `HuggingFaceDownloadManager`, so you have to re-apply filters to find it again.

This surfaces the active download in two always-reachable places:

- **A banner pinned to the top of the Models page** — shown whenever a download is running, above the section content and independent of the section/filters, with the model name, progress, and Pause/Resume + Cancel.
- **A floating chip in the main window chrome** on every other page, showing "Downloading N%", that jumps straight to the Models page when tapped.

Read-only over the download manager — no changes to download logic.

Screenshot 2026-07-25 at 8 40 12 AM

for issue #71


Blaizzy:
Could you share a screenshot?

Lazarus-931:
in the pr description, also here is what it looks like when you leave the Models page:
Screenshot 2026-07-25 at 10 38 52 AM

Blaizzy:
> in the pr description, also here is what it looks like when you leave the Models page:

This is really cool and needed!

Could we contain the update only to models page? maybe show a icon to indicate that we are still downloading to the 
Screenshot 2026-07-25 at 6 31 47 PM

Lazarus-931:
thanks @Blaizzy, i tried and seemed cluttery, and felt that furthermore if if we do put it next to the "Models" nav bar section,  it would conflict if in the future if multiple models are downloaded at the same time. 

i instead tried to make the existing floating section disappearable, will show once build finishes!

PR #102, opened 2026-07-25
Only allow generative (text/vision) models in the chat model picker

Lazarus-931:
## Problem
The chat model picker is a denylist (`LocalModel.isEligibleForLanguageModelPicker` = `!speechToText && !textToSpeech`), so non-generative models a user downloads — encoders like `google-bert/bert-base-uncased` / `FacebookAI/xlm-roberta-base`, embeddings, ASR, time-series — still appear and can be selected as the chat model. Selecting one and starting the server pre-loads it as `text_generation`; mlx-vlm/mlx-lm has no matching architecture, so startup aborts:

```
ERROR - Error loading model google-bert/bert-base-uncased: Model type bert not supported
ModuleNotFoundError: No module named 'mlx_lm.models.bert'
Application startup failed. Exiting.  (exit status 3)
```

The app is then effectively bricked — the server won't come up and the menu bar is unresponsive, with no in-UI way to recover.

## Fix
- **Picker becomes an allowlist of *generative* models:** text **and multimodal vision-language** models both qualify (`.text` *or* `.vision`); only models with no generative capability — encoders, embeddings, ASR, forecasting, pure image-gen — drop out. Multimodal/VLM models are **not** excluded.
- **More robust generative detection:** infer `.text` from causal/generation architectures (`*ForCausalLM` / `*ForConditionalGeneration` / `*LMHeadModel`) in addition to the existing name heuristics, so legitimate chat models (Phi, SmolLM, …) aren't accidentally filtered out.
- **A stale non-generative selection no longer bricks startup:** when the configured chat model's local `config.json` shows only non-generative architectures, its `--model` pre-load flag is dropped (load-on-demand) so the server still starts, with a log note. The check is conservative — any uncertainty leaves the model untouched.

_Screenshots to follow._

### DIFF
--- a/Sources/Nativ/NativModel.swift
+++ b/Sources/Nativ/NativModel.swift
@@ -204,6 +204,18 @@ final class NativModel: ObservableObject {
         var shouldStartMetrics = false
         metricsClient = NativMetricsClient(baseURL: settings.serverBaseURL)
         modelLoadingProgress = settings.normalized().languageModelID == nil ? nil : 0
+        var launchArguments = settings.launchArguments
+        if let languageModelID = settings.normalized().languageModelID,
+           isKnownNonGenerativeModel(languageModelID),
+           let modelFlagIndex = launchArguments.firstIndex(of: "--model"),
+           modelFlagIndex + 1 < launchArguments.count {
+            // A stale, non-chat selection (e.g. a BERT encoder) would make the
+            // server abort while pre-loading it. Start on-demand instead so it
+            // still comes up.
+            launchArguments.removeSubrange(modelFlagIndex...(modelFlagIndex + 1))
+            modelLoadingProgress = nil
+            appendLog("\n\(languageModelID) is not a text-generation model — starting the server without pre-loading it. Pick a chat model to load one.\n")
+        }
         do {
             var launchEnvironment = settings.launchEnvironment
             launchEnvironment["MLX_PLATFORM_ANALYTICS_DB_PATH"] = currentAnalyticsDatabaseURL().path
@@ -211,7 +223,7 @@ final class NativModel: ObservableObject {
                 launchEnvironment[HuggingFaceAuthentication.environmentVariableName] = effectiveHuggingFaceToken
             }
             try server.start(
-                arguments: settings.launchArguments,
+                arguments: launchArguments,
                 environment: launchEnvironment
             )
             isRunning = true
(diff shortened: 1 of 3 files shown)


Blaizzy:
Please double check this doesn't affect images which are supported in the chat version in images tab

Blaizzy:
Also, it doesn't affect omni models like Gemma 4 which might use that tag

Lazarus-931:
yeah @Blaizzy, great point about omni models, added a small patch for that, for the img page, i think we can wait to merge this until a new pr for folding the img feature into the chat page. i think @lucasnewman said it'll be a follow up work

lucasnewman:
> yeah @Blaizzy, great point about omni models, added a small patch for that, for the img page, i think we can wait to merge this until a new pr for folding the img feature into the chat page. i think @lucasnewman said it'll be a follow up work

I added that in https://github.com/Blaizzy/nativ/pull/93 so you should be good -- model availability is type-aware now for non-language models (image, stt, tts).

Sources/Nativ/NativModel.swift
```diff
@@ -236,6 +248,53 @@ final class NativModel: ObservableObject {
...
+
+        var roots: [String] = []
+        let configured = settings.normalized().modelSearchPath.trimmingCharacters(in: .whitespaces)
+        if !configured.isEmpty {
+            roots.append((configured as NSString).expandingTildeInPath)
+        }
+        if let hubCache = ProcessInfo.processInfo.environment["HF_HUB_CACHE"] {
+            roots.append(hubCache)
+        }
+        roots.append(("~/.cache/huggingface/hub" as NSString).expandingTildeInPath)
```
lucasnewman: Does this use the custom roots a user may provide? Should that logic be consolidated so this doesn't have to re-define the roots?

lucasnewman (approved):
LGTM but see comment about making sure to pick up all possible model roots.

Sources/Nativ/NativModel.swift
Lazarus-931: no, but it should be consolidated, i didn't think about non-gen models that only live in custom folders, this would have incorrectly allowed them in the chat. will patch this, thx!

lucasnewman (approved):

Lazarus-931:
> > yeah @Blaizzy, great point about omni models, added a small patch for that, for the img page, i think we can wait to merge this until a new pr for folding the img feature into the chat page. i think @lucasnewman said it'll be a follow up work
> 
> I added that in #93 so you should be good -- model availability is type-aware now for non-language models (image, stt, tts).

wait, confused, the release shows sep img and chat page

PR #104, opened 2026-07-25
Keep the server running when a configured model fails to pre-load

Lazarus-931:
## Problem
An unsupported `model_type` in the configured chat model crashes **server startup**. The server pre-loads `--model` in its lifespan; if mlx-vlm/mlx-lm has no matching architecture, the load raises and the process exits:

    ERROR - Error loading model <id>: Model type <x> not supported
    Application startup failed. Exiting.   (exit status 3)

This happens both for non-generative encoders (`google-bert/bert-base-uncased`) and for models mlx-vlm doesn't implement (`yuriyvnv/WAVe-1B-Multimodal-PT`, `model_type: wave`). The whole app is then effectively bricked — the server won't come up and the menu bar is unresponsive, with no in-UI way to recover.

## Fix
Make a failed pre-load non-fatal. When the server process exits during startup because the configured model failed to load, restart it once **without** pre-loading that model (load-on-demand) and surface a brief notice. Detection lives in the process-termination handler: a non-zero exit + a model-load error seen in the server output + a model was actually configured + it wasn't a deliberate stop — and it retries at most once, so it can't loop. Models that load fine still pre-load normally.

## Result
A bad/unsupported model selection can no longer keep the server — and the app — down: the server comes up and the user can pick another model.

_Screenshots to follow._

### DIFF
--- a/Sources/Nativ/NativModel.swift
+++ b/Sources/Nativ/NativModel.swift
@@ -31,6 +31,7 @@ final class NativModel: ObservableObject {
     @Published private(set) var modelSwitchTargetID: String?
     @Published private(set) var modelLoadingProgress: Double?
     @Published private(set) var modelPreloadMemoryWarning: ModelPreloadMemoryWarning?
+    @Published private(set) var modelPreloadFailureNotice: String?
     @Published private(set) var metricsLoading = false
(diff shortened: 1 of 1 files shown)


Blaizzy:
Closing in favour of #108

PR #107, opened 2026-07-25
Remove the Models sidebar download arrow

Lazarus-931:
Follow-up to #105. That PR both removed the floating download indicator and added a pulsing download arrow to the Models sidebar row. We'd like #105's contribution to stand as just the removal of the floating indicator, so this removes the sidebar arrow (the `ModelsDownloadArrow` view, its `HuggingFaceDownloadManager` observation, and the sidebar usage), restoring the Models row to its prior layout (label + model-loading percentage).

Net effect after #105 + this PR: the floating download indicator is gone, with no sidebar arrow added.


Lazarus-931:
cc: @blaizzy

lucasnewman (approved):

PR #108, opened 2026-07-25
Show model loading failures to users

Blaizzy:
## Summary

- parse structured server error details for chat, image generation, and integrations
- track model startup failures and show dismissible feedback in the Models page, control panel, menu, and session stats
- label unsupported cached models and explain why they cannot be selected when clicked
- add coverage for FastAPI error details, fallback HTTP messages, and startup-log extraction

## Why

mlx-vlm now returns actionable HTTP 400 details when a model cannot be loaded ([Blaizzy/mlx-vlm#1717](https://github.com/Blaizzy/mlx-vlm/pull/1717)), but Nativ did not consistently surface those details. Startup failures could leave the UI looking like it was still loading, and clicking a cached model with no supported preload role was a silent no-op.

## User impact

Users now get the server's useful failure reason in the workflow where loading failed and in app-wide status surfaces. Unsupported models are visibly marked and explain that Nativ cannot use them for its supported serving roles.

## Validation

- `xcodebuild -project Nativ.xcodeproj -scheme Nativ -destination 'platform=macOS,arch=arm64' CODE_SIGNING_ALLOWED=NO test` — 92 tests passed
- `python3 -m py_compile PythonDistribution/Overlay/nativ_server.py PythonDistribution/Scripts/build_mlx_vlm_server.py`
- `git diff --check`

### DIFF
--- a/Sources/Nativ/AppDelegate.swift
+++ b/Sources/Nativ/AppDelegate.swift
@@ -1010,6 +1010,23 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuDelegate {
             return submenu
         }
 
+        if let failure = model.modelLoadFailure {
+            let titleItem = disabledMenuItem(failure.title)
+            titleItem.image = menuIcon(
+                "exclamationmark.triangle.fill",
+                description: "Model loading failed"
+            )
+            titleItem.toolTip = failure.message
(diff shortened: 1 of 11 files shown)


Lazarus-931 (approved):
lgtm!

PR #109, opened 2026-07-25
Remove the circular spinner from the Models download banner

Lazarus-931:
The active-download banner on the Models page shows a small circular progress spinner to the left of the model name and the "Downloading… N%" label. It's redundant with the percentage text (and with the per-row progress ring each model already shows), so this removes it for a cleaner banner. The name, progress label, and Pause/Cancel controls are unchanged.


Blaizzy:
Please share a screenshot of the models page with this change :)

Lazarus-931:
Screenshot 2026-07-25 at 6 42 13 PM

PR #111, opened 2026-07-25
Fix full-screen sidebar corners

Blaizzy:
## Summary

- make the control-panel sidebar flush with the window edges in full screen
- remove the main sidebar glass, backdrop, and layer corner radii while full screen is active
- cover the separate macOS full-screen toolbar glass surface so the top corner is square too
- restore the normal rounded appearance after returning to a window

## Root cause

On macOS 26, AppKit renders the sidebar and the full-screen titlebar glass in separate view and window hierarchies. Adjusting the main sidebar surface fixed the bottom corners, but the independently managed titlebar glass retained its rounded top corners. AppKit also recreates some of these glass layers during the full-screen transition.

## User impact

The sidebar now reaches every full-screen edge with straight corners, while windowed mode keeps its native rounded corners.

## Screenshot

Nativ full-screen sidebar with straight corners

## Validation

- visually verified windowed, full-screen, and full-screen exit/re-entry states
- `xcodebuild -project Nativ.xcodeproj -scheme Nativ -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO test -quiet`
- 92 tests passed, 0 failed


lucasnewman:
fwiw, I don't think this is the right approach -- removing the toolbar breaks a lot of useful behavior we get from SwiftUI and will result in a much jankier feel. what problem are you trying to solve?

Blaizzy:
The issue was the tool bar insets where wrong and nothing fixed it 

Then I started noticing most Mac apps don't use it 

Implement it and did come across some janky feel before fully implementing the new menu system. But now it's working quite smoothly and consistently than ever before.

Please try it and let me know 

image

Sources/Nativ/Main.swift
```diff
@@ -134,11 +134,18 @@ private struct NativRootView: View {
...
                     openWindow(id: "main")
                 }
             }
+            .onChange(of: appearance) { _, newAppearance in
+                applyAppearance(newAppearance)
+            }
+    }
+
+    private func applyAppearance(_ appearance: AppAppearance) {
+        NSApplication.shared.appearance = appearance.appKitAppearance
```
lucasnewman, line 149: This can be handled as a follow up but it should be possible to use `@Environment(\.colorScheme) private var colorScheme` and just directly grab the system color scheme when there's no user override, not need to bounce through AppKit.

PR #117, opened 2026-07-26
Add Buzz as a guided-setup integration

Lazarus-931:
same integration as before, thought it'd best to include it in the readme, but included since we already have blocks' goose. 

env based like claudecode or codex is risky, safe to let user follow instructions, and we can update if things change with buzz


lucasnewman (approved):
LGTM

PR #118, opened 2026-07-26
Add a default image model for the Images view

lucasnewman:
Uses `mlx-community/flux2-klein-4b-8bit` by default, which supports generation and editing and is ~8gb. Trying to generate an image without a model loaded will first download this model and load it, then generate the image.


lucasnewman:
(Note this requires the latest mlx-vlm for the underlying model support)

Blaizzy (approved):
LGTM

PR #119, opened 2026-07-27
Add Mage-Flow family model support

lucasnewman:
This required some additional handling for edit-only models like the Mage-Flow-Edit, hence the hub and capability tweaks.


Blaizzy (approved):
LGTM, thanks!

PR #122, opened 2026-07-27
Store API key in the keychain

### LINKED ISSUE #106: [Security] API key stored in plaintext plist — trivially stealable by any local process
## Summary

The server API key is stored in plaintext in `~/Library/Application Support/Nativ/Settings.plist` using `PropertyListEncoder`. Any process running as the user can read the key with a single command, bypassing the authentication protecting the `/metrics`, `/v1/metrics`, and `/v1/models/load` management endpoints.

## Steps to Reproduce

1. Start Nativ with an API key configured (generated during onboarding)
2. From **any** process (Terminal, another app, a script):

```sh
# Read the key — no privileges needed
plutil -p ~/Library/Application\ Support/Nativ/Settings.plist | grep serverAPIKey
# Output: "serverAPIKey" => "nativ_9XUZoomMp0VrIYjjRDIQ4YD1jWEnfuAqgBEqrKz9CRA"
```

3. Use the stolen key to access protected endpoints:

```sh
# Read detailed metrics — works with stolen key, returns 401 without it
curl -s http://127.0.0.1:8080/metrics \
  -H "Authorization: Bearer <stolen_key>" | python3 -m json.tool

# Switch the loaded model — works with stolen key, returns 401 without it
curl -s -X POST http://127.0.0.1:8080/v1/models/load \
  -H "Authorization: Bearer <stolen_key>" \
  -H "Content-Type: application/json" \
  -d '{"model": "some-other-model"}'
```

4. The key is also passed to the child process via `MLX_VLM_SERVER_API_KEY` environment variable (set at `Sources/Nativ/NativSettings.swift:329`), making it readable through process environment inspection.

## Impact

The API key is the **only** authentication mechanism protecting management endpoints (`/metrics`, `/v1/metrics`, `/v1/models/load`). Since it is stored in plaintext in a well-known location, any local process — a malicious npm postinstall script, a compromised browser extension helper, any app with filesystem access — can:

- Read full request metrics (models used, token counts, endpoint paths, timestamps)
- Switch which model the server is running via `/v1/models/load`
- (Potentially) access any future management endpoints that use this key

Combined with the fact that inference endpoints (`/v1/chat/completions`, `/v1/models`, etc.) require **no authentication at all**, a malicious local process can both use the LLM freely and manipulate the server configuration.

## Root Cause

`Sources/Nativ/NativSettings.swift` stores `serverAPIKey` via the standard `Codable` / `PropertyListEncoder` path to a binary plist. It also passes the key as a plain environment variable to the child Python process.

```swift
// NativSettings.swift:204
try container.encodeIfPresent(serverAPIKey, forKey: .serverAPIKey)

// NativSettings.swift:329
environment["MLX_VLM_SERVER_API_KEY"] = serverAPIKey
```

## Proposed Fix

Store the API key in the macOS Keychain instead of the plist:

```swift
// Write to Keychain
let query: [String: Any] = [
    kSecClass as String: kSecClassGenericPassword,
    kSecAttrService as String: "com.nativ.server-api-key",
    kSecAttrAccount as String: "nativ-server",
    kSecValueData as String: apiKey.data(using: .utf8)!,
    kSecAttrAccessible as String: kSecAttrAccessibleWhenUnlocked
]
SecItemAdd(query as CFDictionary, nil)

// Read from Keychain
let query: [String: Any] = [
    kSecClass as String: kSecClassGenericPassword,
    kSecAttrService as String: "com.nativ.server-api-key",
    kSecAttrAccount as String: "nativ-server",
    kSecReturnData as String: true,
    kSecMatchLimit as String: kSecMatchLimitOne
]
var item: CFTypeRef?
SecItemCopyMatching(query as CFDictionary, &item)
```

The plist should only store a boolean flag indicating whether a key has been generated, not the key itself.

For the child process environment, consider passing the key through a pipe or Unix socket at startup rather than via an environment variable.

## Environment

- Nativ: latest GitHub release
- macOS: 26.5.2
- Hardware: Apple M4 Max

lucasnewman:
This migrates any plist-stored credentials to the keychain as well.

### DIFF
--- a/Sources/Nativ/NativSettings.swift
+++ b/Sources/Nativ/NativSettings.swift
@@ -2,6 +2,90 @@ import Foundation
 import NativServerKit
 import Security
 
+enum ServerAPICredentialPersistenceError: Error {
+    case keychain(OSStatus)
+    case invalidKeychainData
+}
+
+protocol ServerAPICredentialStoring {
+    func load() throws -> String?
+    func save(_ token: String?) throws
+}
+
+struct ServerAPIKeychain: ServerAPICredentialStoring {
+    let service: String
+    let account: String
+
+    init(
+        service: String = "dev.local.Nativ.server-api-key",
+        account: String = "nativ-server"
+    ) {
+        self.service = service
+        self.account = account
+    }
+
+    func load() throws -> String? {
+        var query = baseQuery
+        query[kSecReturnData as String] = true
+        query[kSecMatchLimit as String] = kSecMatchLimitOne
+
+        var item: CFTypeRef?
+        let status = SecItemCopyMatching(query as CFDictionary, &item)
+        if status == errSecItemNotFound {
+            return nil
+        }
+        guard status == errSecSuccess else {
+            throw ServerAPICredentialPersistenceError.keychain(status)
+        }
+        guard let data = item as? Data,
+              let token = String(data: data, encoding: .utf8) else {
+            throw ServerAPICredentialPersistenceError.invalidKeychainData
+        }
+        return ServerAPIAuthentication.normalizedToken(token)
+    }
+
+    func save(_ token: String?) throws {
+        guard let token = ServerAPIAuthentication.normalizedToken(token) else {
+            let status = SecItemDelete(baseQuery as CFDictionary)
+            guard status == errSecSuccess || status == errSecItemNotFound else {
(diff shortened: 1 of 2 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #124, opened 2026-07-27
Add "Sort by Size" to the model selection menu

### LINKED ISSUE #120: Feature Request: Sort by Model Size (Gb) in selection menu
I would like to suggest adding a "Sort by size" option to the model selection interface. Currently, it can be difficult to identify which models are suitable for specific hardware constraints. Providing a sorting option based on model size would allow users to quickly find models that fit their available memory (RAM/VRAM).

Lazarus-931:
Resolves Blaizzy/nativ#120.

Adds a **Size** option to the Discover "Sort by" menu so users can quickly find models that fit their hardware (RAM/VRAM).

**How:** Hugging Face's search API doesn't support sorting by model size, so `.size` fetches by downloads (a valid API sort) and re-sorts the fetched results **client-side** by size — smallest first, with unknown sizes last — using the `safetensors` size metadata the search already requests (`expand[]=safetensors`). Pagination re-orders the accumulated buffer, so paging stays consistent.

**Scope:** one new enum case (`HuggingFaceModelSort.size`) + an `orderedBuffer` in `HuggingFaceModelLibrary`; the sort picker picks it up automatically via `allCases`.

### DIFF
--- a/Sources/Nativ/Features/Models/HuggingFaceHub.swift
+++ b/Sources/Nativ/Features/Models/HuggingFaceHub.swift
@@ -7,6 +7,7 @@ enum HuggingFaceModelSort: String, CaseIterable, Hashable, Identifiable, Sendabl
     case trending = "trendingScore"
     case likes
     case recentlyUpdated = "lastModified"
+    case size
 
     var id: String { rawValue }
 
@@ -16,6 +17,7 @@ enum HuggingFaceModelSort: String, CaseIterable, Hashable, Identifiable, Sendabl
         case .trending: "Trending"
         case .likes: "Likes"
         case .recentlyUpdated: "Recently Updated"
+        case .size: "Size"
         }
     }
 
@@ -25,6 +27,7 @@ enum HuggingFaceModelSort: String, CaseIterable, Hashable, Identifiable, Sendabl
         case .trending: "flame"
         case .likes: "heart"
         case .recentlyUpdated: "clock.arrow.circlepath"
+        case .size: "internaldrive"
         }
     }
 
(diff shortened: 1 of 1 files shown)


Sources/Nativ/Features/Models/HuggingFaceHub.swift
```diff
@@ -34,8 +37,18 @@ enum HuggingFaceModelSort: String, CaseIterable, Hashable, Identifiable, Sendabl
...
         case .recentlyUpdated: "modified"
+        case .size: "downloads"
         }
     }
+
+    /// The Hugging Face search API can't sort by model size, so `.size` fetches by
+    /// downloads and the results are re-sorted client-side.
+    var apiValue: String {
+        self == .size ? Self.downloads.rawValue : rawValue
+    }
```
Blaizzy: @lucasnewman could you check this change?

Blaizzy (commented):
LGTM, but I have one part where the comment is a bit odd, so @lucasnewman will double check.

Sources/Nativ/Features/Models/HuggingFaceHub.swift
```diff
@@ -322,7 +335,7 @@ private struct HuggingFaceHubClient: Sendable {
 
         var queryItems = [
             URLQueryItem(name: "filter", value: "safetensors"),
-            URLQueryItem(name: "sort", value: sort.rawValue),
+            URLQueryItem(name: "sort", value: sort.apiValue),
```
Blaizzy: Revert

Sources/Nativ/Features/Models/HuggingFaceHub.swift
```diff
@@ -510,9 +525,24 @@ final class HuggingFaceModelLibrary: ObservableObject {
...
         let start = (number - 1) * pageSize
-        guard start < buffer.count else { return [] }
-        return Array(buffer[start..<min(start + pageSize, buffer.count)])
+        guard start < ordered.count else { return [] }
+        return Array(ordered[start..<min(start + pageSize, ordered.count)])
+    }
+
+    /// The buffered results in display order. Hugging Face can't sort by size, so the
+    /// `.size` sort re-orders the fetched results locally (smallest first; models whose
+    /// size is unknown go last).
```
Blaizzy: short comment

Blaizzy (approved):

PR #126, opened 2026-07-27
Support concurrent model downloads

Lazarus-931:
Replace the single-download HuggingFaceDownloadManager with a queue that runs up to a user's device suffices. Each download tracks its own operation, task, progress, state (downloading/paused/queued) and waiters, keyed by model ID; extra requests are queued FIFO and started as active slots free up on completion or cancel. Pause (SIGSTOP), resume (SIGCONT) and cancel now target an individual model.

Add a cumulative disk gate: capacityBlocker compares a model's size against free space minus reservedBytes (the remaining bytes of every in-flight download), so a download is only offered when it still fits.

Update all consumers: the sidebar arrow shows the active count, the Models banner lists one row per download with per-model controls, and hub rows drive their state per model while disabling downloads that no longer fit on disk.

### DIFF
--- a/Sources/Nativ/ControlPanelView.swift
+++ b/Sources/Nativ/ControlPanelView.swift
@@ -117,17 +117,30 @@ extension Color {
 /// A small pulsing download arrow shown at the trailing edge of the Models sidebar row
 /// while a model is downloading.
 private struct ModelsDownloadArrow: View {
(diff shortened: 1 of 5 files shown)


Blaizzy (approved):
LGTM, thanks!

Issue #130, opened 2026-07-27
Allow downloads to proceed in the background of the nativ setup


Lazarus-931:
Just saw this but users shouldn't need to wait for models to download when first setting up nativ, should be prompted to continue and let download process run in background: 

<img width="860" height="657" alt="Image" src="https://github.com/user-attachments/assets/22fd0ea7-54b5-4b03-a9b2-329b96f20b3d" />

lucasnewman:
Resolved in https://github.com/Blaizzy/nativ/pull/132

PR #132, opened 2026-07-27
Allow setup to continue while model downloads run in the background

Lazarus-931:
When user is setting up nativ, the app does not allow that person to continue while recommended model is downloading, this pr patches that, allows them to continue to nativ.

patches Blaizzy/nativ#130

### DIFF
--- a/Sources/Nativ/Features/Models/HuggingFaceHub.swift
+++ b/Sources/Nativ/Features/Models/HuggingFaceHub.swift
@@ -350,6 +350,35 @@ private struct HuggingFaceHubClient: Sendable {
         return try await page(at: url, token: token)
     }
 
+    func model(id: String, token: String?) async throws -> HuggingFaceModel {
+        var components = URLComponents()
+        components.scheme = "https"
+        components.host = "huggingface.co"
+        components.path = "/api/models/\(id)"
+        components.queryItems = [
+            "downloads", "likes", "pipeline_tag", "library_name", "tags",
+            "private", "gated", "safetensors"
+        ].map { URLQueryItem(name: "expand[]", value: $0) }
+
+        guard let url = components.url else {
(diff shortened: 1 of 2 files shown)


Blaizzy:
Let's take a different approach, lets make the default a small model like LFM because if the model is not finished by the time they finish onboarding it breaks the flow.

Blaizzy:
Lets pick 3-6 small models and make those show up on the list

VLMs and LMs

Lazarus-931:
sounds good, I think we should have a wide variety of options for diff hardware scales, i recommend(from smallest to biggest):

- Qwen3-VL-4B
- Qwen3.5-9B
- Qwen3.6-27B
- GLM-4.7-Flash
- Qwen3-VL-32B
- gpt-oss-120b

Lazarus-931:
@Blaizzy, updated the code so that now these following models are recommended:

LFM2.5-VL-450M 
LFM2.5-1.2B 
LFM2.5-VL-1.6B 
Qwen3-VL-4B 
Qwen3.5-9B
GLM-4.7-Flash
Qwen3-VL-32B
gpt-oss-120b

Furthermore, it makes sure to reuse the component to prevent over-downloading several models beyond device capability.

but i think a continue option while downloading is still needed because for me at least even the smallest two take up to 3-4 mins of download time.

Lazarus-931:
cc: @Blaizzy

lucasnewman:
@Lazarus-931 It will show the model downloading banner in chat, right? If so, I think it's fine (actually better / less frustrating) to allow continuing from setup.

lucasnewman (approved):

Lazarus-931:
Changed it to:

https://github.com/user-attachments/assets/08874916-5671-4351-821d-3de3c2b157bf

PR #133, opened 2026-07-27
Keep selection menus open

Blaizzy:
## What changed

- Keep the chat model and reasoning menus open after selecting an option.
- Keep the menu bar customization menu open while toggling metrics and display styles.
- Update checkmarks and selected-value summaries in place.
- Share the persistent AppKit menu-row behavior through `PersistentMenuActionView`.

## Why

The model and menu bar customization pickers previously used ordinary menu actions. AppKit and SwiftUI ended menu tracking after every selection, forcing users to reopen the menu for each additional choice.

This change uses embedded menu controls that apply the selection without ending the active menu session. The menus still dismiss through the normal outside-click or cancellation behavior.

## User impact

Users can switch models or reasoning levels and configure several menu bar items without repeatedly reopening the relevant menu.

## Validation

- Debug macOS build succeeds.
- Full test suite passes: 108 tests, 0 failures.
- `git diff --check` passes.


lucasnewman (approved):
LGTM

PR #138, opened 2026-07-28
Make Return activate the default button in confirmation dialogs

Lazarus-931:
One fearture is to allow the submit button to work on like small pop ups that show, such as these:

I think it's just something natural most people do, so would be nice
Screenshot 2026-07-28 at 7 44 33 AM


lucasnewman (approved):
👍

PR #140, opened 2026-07-28
Fix settings not persisting (Model Configuration resets, model source folders lost)

### LINKED ISSUE #135: Additional model sources folder not persisting across app restart
Version 0.1.0 (202607272042) + OS X 26.5.2. Prebuilt version downloaded from https://blaizzy.github.io/nativ/

Steps to reproduce: 
1. Model Sources -> Add Folder... 
2. After restarting the app is back to just the  default folder.

Add folder, model detection works ok:
Image

After restarting app:
Image

Let me know if I can do anything to help!

### LINKED ISSUE #139: Model Configuration resets and is not persisted
## Environment

- Nativ version: 0.1.0
- Build: 202607272042
- Installed from the official DMG in `/Applications/Nativ.app`
- macOS: 26

## Steps to reproduce

1. Select/load a language model in **Models**.
2. Open **Chat** → **Model Configuration**.
3. Change **Temperature** from `0` to `0.1`.
4. Close and reopen the Model Configuration panel.

## Expected

The changed value remains saved, at least for the current model or app session.

## Actual

The value resets to `0`. Other Model Configuration values also reset.

## Storage checks

- `~/Library/Application Support/Nativ/Settings.plist` is not created.
- `~/Library/Preferences/io.github.blaizzy.nativ.plist` contains only window, updater, and welcome settings—not model configuration.
- `~/Library/Application Support/Nativ/` is writable.
- No Nativ crash report was created.

Lazarus-931:
Closes Blaizzy/nativ#139. Closes Blaizzy/nativ#135.

## Problem

Settings changes don't survive an app restart — Model Configuration (temperature, etc.) resets, and added model source folders disappear. As reported in Blaizzy/nativ#139, `~/Library/Application Support/Nativ/Settings.plist` is **never created**, even though the directory is writable.

## Cause

`NativSettings.save()` wrote the keychain **before** the plist, inside one `do/catch` that swallows errors:

```swift
let settings = normalized()
do {
    try credentialStore.save(settings.serverAPIKey)   // keychain first
    try settings.writePropertyList(to: url)           // skipped if the line above throws
} catch { /* swallowed */ }
```

The keychain query uses `kSecUseDataProtectionKeychain: true`. On a non-sandboxed app without a keychain-access-groups entitlement (which this app is), that call fails, so `credentialStore.save()` throws, `writePropertyList` never runs, the error is eaten, and no settings are ever written to disk. This affects every setting — Model Configuration (Blaizzy/nativ#139) and model source folders (Blaizzy/nativ#135) alike.

## Fix

Write the settings plist independently of the keychain, so a keychain failure can't block it:

```swift
let settings = normalized()
try? settings.writePropertyList(to: url)          // always persist settings
try? credentialStore.save(settings.serverAPIKey)  // best-effort key storage
```

`serverAPIKey` is not encoded into the plist (it's only read for legacy migration), so writing the plist doesn't leak the key to disk. The key still reaches the keychain when that succeeds; when it doesn't, settings persist anyway and the key is simply session-only — strictly better than today, where nothing persists.

## Testing

Change Temperature, or add a model source folder, then quit and relaunch → the value/folder is retained, and `Settings.plist` is created.

## Follow-up

Dropping `kSecUseDataProtectionKeychain: true` would let the keychain write itself succeed on this app (so the API key persists across launches too). Left out here to keep this change minimal and avoid keychain-migration concerns.

### DIFF
--- a/Configuration/Nativ.entitlements
+++ b/Configuration/Nativ.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>keychain-access-groups</key>
+	<array>
+		<string>$(AppIdentifierPrefix)$(PRODUCT_BUNDLE_IDENTIFIER)</string>
+	</array>
+</dict>
+</plist>
--- a/Sources/Nativ/NativSettings.swift
+++ b/Sources/Nativ/NativSettings.swift
@@ -573,12 +573,8 @@ struct NativSettings: Codable, Equatable {
         credentialStore: ServerAPICredentialStoring = ServerAPIKeychain()
     ) {
         let settings = normalized()
-        do {
-            try credentialStore.save(settings.serverAPIKey)
-            try settings.writePropertyList(to: url)
-        } catch {
-            // Do not replace the last recoverable credential on failure.
-        }
+        try? settings.writePropertyList(to: url)
+        try? credentialStore.save(settings.serverAPIKey)
     }
 
     private func writePropertyList(to url: URL) throws {
--- a/Tests/NativTests/NativSettingsTests.swift
+++ b/Tests/NativTests/NativSettingsTests.swift
@@ -148,6 +148,25 @@ final class NativSettingsTests: XCTestCase {
         )
     }
 
+    func testSavingWritesPropertyListEvenWhenCredentialStoreFails() throws {
+        let url = temporarySettingsURL()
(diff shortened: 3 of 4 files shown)


lucasnewman:
@Lazarus-931 We should fix the keychain param first, ideally by adding the entitlement before dropping the param. It's trying to avoid losing data since the settings write will drop the migrated API key.

C-K-Loan:
FYI, #143 fixes the same underlying bug (`NativSettings.save()`'s Keychain write and plist write sharing one swallowing `do/catch`) — found independently while testing an unrelated feature, so this is a nice cross-confirmation that the root cause and the fix shape are right. Didn't know about #139/#135 when we filed ours; thanks for tracking those down. #143 does add a regression test pinning the behavior (`testSavingWritesPropertyListEvenWhenCredentialStoreFails`) if that's useful to fold in regardless of which PR ends up landing.

lucasnewman (approved):
🙏

PR #142, opened 2026-07-28
View > Toggle Sidebar was permanently disabled

C-K-Loan:
The View menu's "Show Sidebar" command and its ⌃⌘S keyboard shortcut have never worked — `SidebarCommands()` is SwiftUI's stock API for a real `NavigationSplitView`, but this app's sidebar is a custom `ZStack` and has never had one. The menu item read `enabled: false` unconditionally.

This wires the menu command to the app's actual sidebar-visibility state, mirroring the existing `newChatRequest` pattern used elsewhere in the app. The toolbar sidebar-toggle button (a separate, custom-built control) is unaffected — this only fixes the standards-based menu/keyboard path.

Small, self-contained fix — unrelated to the tool-registry work in the other open PR.

### DIFF
--- a/Sources/Nativ/AppDelegate.swift
+++ b/Sources/Nativ/AppDelegate.swift
@@ -504,6 +504,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuDelegate {
         showMainWindow()
     }
 
+    func toggleSidebar() {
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
@C-K-Loan Thanks!

PR #143, opened 2026-07-28
Settings never persisted across a relaunch on unsigned debug builds

C-K-Loan:
`NativSettings.save()` ran the Keychain write and the property-list write in the same `do`/`catch` block. On an unsigned or ad-hoc-signed build (no `keychain-access-groups` entitlement), the Keychain write throws `errSecMissingEntitlement`, which silently aborted the plist write too — meaning **no settings change survived an app relaunch** in that build configuration, not just newly-added items.

Likely low-impact for properly signed/notarized distribution builds (which do have the entitlement), but real for anyone building and running Nativ locally without signing — which is presumably common for contributors.

Fix: decouple the two writes so a recoverable-credential failure can't block the rest of settings from being written to disk.

Found and fixed while testing an unrelated feature; small, self-contained, unrelated to the other open PRs.

### DIFF
--- a/Sources/Nativ/NativSettings.swift
+++ b/Sources/Nativ/NativSettings.swift
@@ -575,10 +575,10 @@ struct NativSettings: Codable, Equatable {
         let settings = normalized()
         do {
             try credentialStore.save(settings.serverAPIKey)
-            try settings.writePropertyList(to: url)
         } catch {
             // Do not replace the last recoverable credential on failure.
         }
+        try? settings.writePropertyList(to: url)
     }
 
     private func writePropertyList(to url: URL) throws {
--- a/Tests/NativTests/NativSettingsTests.swift
+++ b/Tests/NativTests/NativSettingsTests.swift
@@ -148,6 +148,25 @@ final class NativSettingsTests: XCTestCase {
         )
     }
 
+    func testSavingWritesPropertyListEvenWhenCredentialStoreFails() throws {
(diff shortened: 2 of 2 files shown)


C-K-Loan:
Just noticed #140 fixes the exact same root cause (`NativSettings.save()` coupling the Keychain write and the plist write in one swallowing `do/catch`) — independently found while testing an unrelated feature, so this is genuine cross-confirmation of the bug and the fix shape, not a copy. #140 additionally closes #139 and #135, which I hadn't seen. This PR does carry a regression test (`testSavingWritesPropertyListEvenWhenCredentialStoreFails`) pinning the behavior, which might be worth folding into whichever version lands. Happy to close this in favor of #140, or let the maintainer pick — just flagging the overlap.

Lazarus-931:
> Just noticed #140 fixes the exact same root cause (`NativSettings.save()` coupling the Keychain write and the plist write in one swallowing `do/catch`) — independently found while testing an unrelated feature, so this is genuine cross-confirmation of the bug and the fix shape, not a copy. #140 additionally closes #139 and #135, which I hadn't seen. This PR does carry a regression test (`testSavingWritesPropertyListEvenWhenCredentialStoreFails`) pinning the behavior, which might be worth folding into whichever version lands. Happy to close this in favor of #140, or let the maintainer pick — just flagging the overlap.

hi @C-K-Loan, thanks for the pr, yes #140 is open and I'd recommend closing this!

PR #144, opened 2026-07-28
Include downloaded image models in the Models installed list

Lazarus-931:
Flux model downloaded for img gen page doesn't show in Models page, under the list of installed models, this pr fixes that. By this, people do have the choice to delete the model if need be.


lucasnewman (approved):
LGTM

PR #146, opened 2026-07-29
Add pinned chats: Pinned and Sessions sidebar sections with a chat co…

Lazarus-931:
Sidebar looks like this now, cleaner:

Screenshot 2026-07-28 at 10 10 58 PM

### DIFF
--- a/Sources/Nativ/ControlPanelView.swift
+++ b/Sources/Nativ/ControlPanelView.swift
@@ -2417,6 +2417,8 @@ private struct ControlPanelRecentSessionRow: View {
                 Button {
                     if NSEvent.modifierFlags.contains(.shift), recent.isChat {
                         onTogglePin()
+                    } else if isSelected, recent.isChat {
+                        beginRename()
                     } else {
                         onSelect()
                     }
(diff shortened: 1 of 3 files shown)


Sources/Nativ/ControlPanelView.swift
```diff
@@ -2369,30 +2477,46 @@ private struct ControlPanelRecentSessionRow: View {
         .opacity(isSelectionDisabled && !isCurrent ? 0.55 : 1)
         .onHover { isHovering = $0 }
         .animation(.easeInOut, value: isHovering)
+        .simultaneousGesture(
+            TapGesture(count: 2).onEnded {
```
lucasnewman: This is minor but the most standard rename interaction on macOS is a single click on an already-selected row instead of a double-click. You can see what I mean in the sidebar of e.g. Xcode. I realize Codex diverges from this so I don't have a super strong opinion either way, but long-time Mac users generally appreciate the consistency of the standard interaction.

lucasnewman (approved):
Nice!

Sources/Nativ/ControlPanelView.swift
Lazarus-931: not a problem, will make it single click,thx!

PR #151, opened 2026-07-30
Add drag-and-drop pinning, reordering, and bulk actions to the chat s…

Lazarus-931:
Replace shift-click pinning with drag-and-drop for pinning, unpinning, and reordering pinned chats, backed by a persisted pinnedOrder so the arrangement survives restarts. Add a per-row actions menu (shared with the right-click menu) and a multi-select mode for bulk pin, export, and delete.

edit: video below

https://github.com/user-attachments/assets/6eee77d1-b1ce-4aba-bfa1-c27a16808357

### DIFF
--- a/Sources/Nativ/ControlPanelView.swift
+++ b/Sources/Nativ/ControlPanelView.swift
@@ -234,6 +234,7 @@ struct ControlPanelView: View {
     @State private var isPinnedDropTargeted = false
     @State private var isSessionsDropTargeted = false
     @State private var reorderTargetID: ControlPanelRecentSession.ID?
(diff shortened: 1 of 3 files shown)


lucasnewman (approved):
LGTM. The ControlPanelView is getting large, we might consider breaking that into subcomponents / files if they are clearly separable.

Lazarus-931:
yes, i agree I was thinking spacing/font size was making that area look cluttery/ too much

lucasnewman:
> yes, i agree I was thinking spacing/font size was making that area look cluttery/ too much

I meant at a code-level fwiw

PR #153, opened 2026-07-30
Don't bundle training-only Python dependencies

Lazarus-931:
Some mlx-vlm deps aren't needed for any reason here, dropping those gives up  ~200 mb free

for #26


lucasnewman (approved):
👌

PR #155, opened 2026-07-30
Fix Japanese and Chinese IME composition in chat input

Blaizzy:
## What changed

- Let the native NSTextView handle key events while an input method has marked text.
- Preserve the existing Return-to-send and Command-Return newline behavior after composition is committed.

## Why

The chat composer intercepted Return before Japanese and Chinese input methods could confirm their marked composition, causing candidate selection to trigger send instead of committing the composed text.

## Validation

- App build succeeded with Xcode 26.6.
- All 147 unit tests passed.
- git diff --check passed.
- Built app opened successfully with the local server on 127.0.0.1:8081 for isolated UI testing because port 8080 was already in use.
- Japanese input: こんにちは、これは日本語の入力テストです。  
  English translation: “Hello, this is a Japanese input test.”  
  Result: Rendered correctly in the chat input box and enabled Send.
- Chinese input: 你好，这是中文输入测试。  
  English translation: “Hello, this is a Chinese input test.”  
  Result: Rendered correctly in the chat input box and enabled Send.

### UI captures

Japanese input:

Japanese CJK chat input validation

Chinese input:

Chinese CJK chat input validation

The UI validation entered committed Unicode text for both scripts. The marked-composition handling is implemented in ChatComposer.


lucasnewman (approved):
Thanks!

PR #156, opened 2026-07-30
Fix Discover sort-by-size (invalid Hugging Face API parameter)

Lazarus-931:
Sorting the Discover list by **Size** sent `sort=size` to the Hugging Face API, which rejects it (`Invalid sort parameter: size`) — so the list showed **0 results** with a "Hugging Face is unavailable" error.

The Hub API has no `size` sort; results are already re-sorted client-side by size (`orderedBuffer`, gated on `sortsBySize`). This sends a valid ordering (`downloads`) for the Size option and lets the existing client-side re-sort produce the size ordering.

I used to get this:
Screenshot 2026-07-30 at 4 51 37 PM


lucasnewman (approved):
LGTM

PR #157, opened 2026-07-31
Add folder upload to the chat composer

Lazarus-931:
This adds support for folder uploads in the chat page. This is important in case people want to upload entire folders or code bases to the model. It is respective of the model's context size, so the bigger the model, the more tokens or the more it's able to read.

This also renames it to Upload File since not just images are supported


lucasnewman:
I think "upload" might be the wrong verb here -- the data isn't leaving the device and I don't think we want to give the impression that it is.

If we had tools to inspect the filesystem, we could just provide the underlying path instead of collating all the data, right? I wonder if that's a more feasible path vs trying to jam an arbitrary amount of data into the context window, and I believe that's what e.g. Codex does today.

Blaizzy:
+1

Lazarus-931:
like shell tools or forcing it to execute inside a specific folder?

Blaizzy:
Any updates here?

PR #159, opened 2026-07-31
Add chat session folders to organize the sidebar

Lazarus-931:
PR for being able to make session folders to organize different chat sessions into

Looks like:

https://github.com/user-attachments/assets/4ff2f7ac-ad29-46b9-8af2-878c2b8c2ab5

### DIFF
--- a/Sources/Nativ/ControlPanelView.swift
+++ b/Sources/Nativ/ControlPanelView.swift
@@ -555,7 +555,7 @@ struct ControlPanelView: View {
     private var pinnedSection: some View {
         VStack(alignment: .leading, spacing: 0) {
             sidebarPinnedHeader
-                .padding(.leading, 17)
+                .padding(.leading, 8)
                 .padding(.trailing, 10)
                 .padding(.bottom, 4)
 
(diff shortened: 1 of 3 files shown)


lucasnewman:
@Lazarus-931 Looks cool! A couple of things I noticed:
1) The blue underline should drop away after a drag, it's kind of unclear why it's there sometimes and not sometimes in your video.
2) It looks like the layout is shifting on hover -- if possible, the row sizes should stay fixed and the hover just shows the effect behind.
3) The empty pin button feels busy. I think it might be nicer to only show the pin icon when a thread is pinned, and then either reveal the empty icon on hover or just use the menu action to keep the interface clean. EDIT: I just noticed you entered a pinning "mode" which is why the icons are there. I think it's actually better/simpler without the mode and with the other changes I mentioned.

Lazarus-931:
thanks @lucasnewman, I think those are valid, will patch!

Lazarus-931:
@lucasnewman, now looks like:

https://github.com/user-attachments/assets/3bee1034-e31d-4a8b-9069-7fed3ecc80b0

lucasnewman:
Looking good!

lucasnewman (approved):

PR #162, opened 2026-07-31
Semantic search in the artifacts page (embedding endpoint)

Lazarus-931:
Stacked on top of #158 (the Artifacts page) — intended to merge **after** #158.

Adds opt-in **semantic search** to the artifacts page, powered by the local mlx-vlm `/v1/embeddings` endpoint:

- A small embedding model (`arthurcollet/Qwen3-VL-Embedding-2B-mlx-4bit`, ~1.8 GB) is offered as an opt-in download; searching by content only kicks in once it's installed, otherwise search falls back to the existing substring match.
- Content search spans **image content**, **video keyframes** (4 frames/clip), and **document text** (PDF via PDFKit + plain-text files).
- Dual-space matching: image/query uses a caption prompt, text/query uses a retrieval prompt, each with its own threshold; an artifact matches if either space clears its bar. Indexing is lazy (only on a query) and cancellable.

Note: requires an mlx-vlm build with the image-embedding `/v1/embeddings` path + quantized-embedding loader fix;

needs #171 and https://github.com/Blaizzy/mlx-vlm/pull/1773 merged


Sources/Nativ/Features/Artifacts/ArtifactsView.swift
```diff
@@ -128,9 +161,255 @@ struct ArtifactsView: View {
...
+        if let config = semanticSearch, !config.isModelInstalled, !config.isDownloading,
+           config.canInstall, !semanticSearchOffered, !store.artifacts.isEmpty {
+            HStack(spacing: 12) {
+                Image(systemName: "sparkle.magnifyingglass")
+                    .font(.system(size: 16))
+                    .foregroundStyle(.secondary)
+                VStack(alignment: .leading, spacing: 2) {
+                    Text("Turn on Smart search")
+                        .font(.system(size: 12, weight: .semibold))
+                    Text("Install a \(config.sizeLabel) on-device model to find artifacts by what's inside them. You can also do this later from the gear menu.")
```
lucasnewman: We should probably say "settings" not "gear menu" for clarity.

Sources/Nativ/ControlPanelView.swift
```diff
@@ -268,6 +269,63 @@ struct ControlPanelView: View {
     @StateObject private var systemMonitor = SystemMonitorStore()
     @StateObject private var launchAtLogin = LaunchAtLoginController()
     @ObservedObject private var downloads = HuggingFaceDownloadManager.shared
+    @StateObject private var embeddingLibrary = LocalModelLibrary()
+
+    private static let embeddingModelID = "mlx-community/Qwen3-VL-Embedding-2B-bf16"
```
lucasnewman, line 274: Should this be configurable? Maybe the answer is "not yet" which is fine, I'm just curious about the reason.

Sources/Nativ/Features/Artifacts/ArtifactsView.swift
```diff
@@ -128,9 +161,255 @@ struct ArtifactsView: View {
...
+        guard smartSearchEnabled, let config = semanticSearch, config.isModelInstalled else {
+            semanticMatches = nil
+            return
+        }
+        let query = search.trimmingCharacters(in: .whitespacesAndNewlines)
+        guard !query.isEmpty else {
+            semanticMatches = nil
+            return
+        }
+        searchDebounce = Task {
```
lucasnewman, line 271: This task is bound to the lifecycle of the view fwiw. Maybe it should be indexing independently with `Task.detached` so the index is prepared prior to searching?

Sources/Nativ/ControlPanelView.swift
```diff
@@ -268,6 +269,63 @@ struct ControlPanelView: View {
...
+        let settings = model.settings.normalized()
+        let baseURL = URL(string: "http://127.0.0.1:\(settings.serverPort)")
+            ?? URL(string: "http://127.0.0.1:8080")!
+        let modelID = Self.embeddingModelID
+        let hasEnoughMemory = ProcessInfo.processInfo.physicalMemory >= 16_000_000_000
+        let diskBlocker = downloads.capacityBlocker(
+            sizeBytes: Self.embeddingModelSize,
+            cachePath: settings.modelSearchPath
+        )
+        let insufficientReason: String? = !hasEnoughMemory
```
lucasnewman: Should we even advertise the feature if the user has < 16GB of RAM? That's not something they can easily change...

Sources/Nativ/ControlPanelView.swift
Lazarus-931: my move with this is the embedding model is pre-selected for now, and in the future, i definitely think this will be a option, this model is just the best for it's weight and what it provides. 

tested several and this one was best

PR #164, opened 2026-08-01
Show a compact memory-fit warning for oversized models

Lazarus-931:
From:

image

to:

Screenshot 2026-08-01 at 7 56 17 AM

Didn't make sense that i'd say "May not fit" when it definitely can't


lucasnewman (approved):
Looks good!

Issue #160, opened 2026-07-31
Per-model configuration profiles — settings like thinking and drafter are not remembered per model


maya528:
Issue 1

Category
Feature request
What happened
Chat model configuration is global, not per-model. There is no way to save a configuration profile for a specific model, which causes two problems:
1. Thinking preference is overwritten on every model switch.
When switching models in the UI, the app unconditionally resets the thinking setting based on the model's capabilities: reasoning-capable models are forced to thinking=max, non-reasoning models to off. If I manually disable thinking for a reasoning model, switch away, and switch back, thinking is re-enabled — my choice is lost.
2. Drafter configuration leaks across models.
"Enable drafter" and the draft model persist globally. A drafter is usually paired with one specific main model (e.g., an MTP drafter for a specific Qwen model). After switching to another model, the previous drafter config stays active and applies to the wrong model, so I have to manually fix it every time.
Steps to reproduce (thinking):
Load a reasoning-capable model.
In Model Configuration, disable "Enable Thinking".
Switch to another model, then switch back.
Thinking is re-enabled — the manual setting was not preserved.
Steps to reproduce (drafter):
Load model A, enable drafter with A's matching draft model.
Switch to model B (no compatible drafter).
Model A's drafter configuration is still active.
Expected behavior
Each model remembers its own configuration. Switching back to a model restores the settings last used with it — at minimum: thinking on/off (+budget), drafter on/off, draft model, and draft kind.
Suggested precedence when a model is loaded:
If the user has explicitly configured this model before → restore the saved per-model values.
Otherwise → apply the current capability-based defaults (reasoning → thinking=max, non-reasoning → off).
A minimal implementation could store a small per-model overrides dictionary keyed by model ID, layered on top of the existing global settings as defaults for new models.
Related
#<ISSUE-2> — applies the same per-model configuration when a model is selected via the API. That issue depends on the per-model store proposed here, but this UI-side feature is useful on its own.
<details>
<summary>Diagnostics</summary>

Environment
Nativ: 0.1.0 (202607272042)
macOS: macOS 26.5.2 (Build 25F84)
Chip: Apple M1 Pro
Memory: 32 GB total
mlx-vlm: 0.6.8
</details>

Lazarus-931:
[...]

PR #166, opened 2026-08-01
Per-model configuration profiles

Lazarus-931:
Addresses #160, creating a small ModelConfigProfile (thinking on/off + budget, drafter on/off + draft model/kind) that is stored per model ID in NativSettings.modelConfigs and persisted to disk. On model switch, ChatComposer snapshots the outgoing model's settings and restores the incoming model's saved profile (or a capability default with the drafter cleared for never-configured models), and a single observer saves any edit immediately so choices stick. Added a guard that re-disables thinking on models that can't do it, plus five unit tests on the store logic.


lucasnewman (approved):
Nice!

PR #167, opened 2026-08-01
Apply per-model configuration on the API path

Lazarus-931:
This PR makes the server apply a model's own saved configuration when that model is named in an API request, instead of inheriting whatever settings were last active globally. On the server side, a new middleware step reads the requested model and fills in its saved enable_thinking/thinking_budget from a per-model config map whenever the client omits them

tested it, and on a live endpoint, when a request names a model, the server now applies that model's saved thinking setting when the client omits it, an explicit client value always overrides it, unrelated models are left alone, and a missing/broken config safely falls back to normal behavior.

this is for text models, but if needed for other model types, can open a pr for that as well; such as img-gen knobs like steps, guidance, size, sampler


lucasnewman (approved):
👍

PR #168, opened 2026-08-01
Warn when attaching images to a non-vision model

Lazarus-931:
Looks like this:

Screenshot 2026-08-01 at 10 55 52 AM


lucasnewman (approved):
LGTM

PR #174, opened 2026-08-02
Chat: drag files from Finder + paste images into the composer

Lazarus-931:
Drag image files from Finder onto the chat composer to stage them as attachments (with a drop highlight).


lucasnewman (approved):
🙏

PR #175, opened 2026-08-03
Improve page responsiveness and background work

Blaizzy:
## Summary

- coalesce and briefly cache duplicate local-model scans
- reduce SwiftUI chat row work and stream flush frequency
- isolate chat transcript and composer updates from the parent chat page
- load chat history off the main actor during initial page construction
- throttle audio meter UI publications while preserving smoothing
- isolate high-frequency audio meter state so it does not refresh the full Audio page
- cache audio analytics aggregates and reuse one SQLite connection/read snapshot for dashboard reloads
- avoid unnecessary model/transcript work when switching audio pages
- refresh integration statuses concurrently instead of serially waiting on each executable

## Benchmarks

Compared with `origin/main` (`4028b46`) using identical local synthetic workloads:

| Workload | `origin/main` | Optimized | Improvement |
| --- | ---: | ---: | ---: |
| 100 repeated local-model scans | 70.3 ms | 0.77 ms | ~92x faster |
| 40 dashboard read batches | 69.1 ms | 5.8 ms | ~12x faster |

The model-scan benchmark exercises cache/coalescing. The dashboard benchmark exercises connection reuse across six analytics reads per reload. Interactive SwiftUI/audio frame timing still needs Instruments on a production device; `xctrace` was unavailable in the benchmark environment.

## Validation

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project Nativ.xcodeproj -scheme Nativ -configuration Debug -derivedDataPath build/NativDevelopmentDerivedData CODE_SIGNING_ALLOWED=NO build`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild -project Nativ.xcodeproj -scheme Nativ -configuration Debug -derivedDataPath build/NativDevelopmentDerivedData CODE_SIGNING_ALLOWED=NO test` (207 tests passed)

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatView.swift
+++ b/Sources/Nativ/Features/Chat/ChatView.swift
@@ -12,63 +12,27 @@ struct ChatQueuedPrompt: Identifiable, Equatable {
     let position: Int
 }
 
-struct ChatView: View {
-    private enum Layout {
-        static let conversationMaxWidth: CGFloat = 680
-        static let horizontalPadding: CGFloat = 32
-    }
+private struct ChatSessionBootstrap {
+    let sessions: [ChatSession]
+}
 
+struct ChatView: View {
     @ObservedObject var model: NativModel
-    @ObservedObject var chat: ChatViewModel
+    let chat: ChatViewModel
     @Binding var showsConfiguration: Bool
     let conversationWidthReduction: CGFloat
-    @State private var transcriptScrollPosition = ScrollPosition(edge: .bottom)
-    @State private var composerHeight: CGFloat = 0
     @State private var isDropTargeted = false
-    @State private var followsLatestMessage = true
-    @State private var isUserScrollingTranscript = false
 
     var body: some View {
         ModelConfigurationLayout(
             model: model,
             isConfigurationVisible: $showsConfiguration
         ) {
-            VStack(spacing: 0) {
-                transcript
-                    .overlay(alignment: .bottom) {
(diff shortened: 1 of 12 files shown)


Sources/Nativ/Features/Chat/ChatView.swift
```diff
@@ -12,92 +12,52 @@ struct ChatQueuedPrompt: Identifiable, Equatable {
...
-            ?? model.settings.structuredOutputValidationError
+    private var selectedModelID: String? {
+        model.settings.normalized().languageModelID
     }
 
-    private var transcript: some View {
+    var body: some View {
         ScrollView {
-            VStack(alignment: .leading, spacing: 12) {
+            LazyVStack(alignment: .leading, spacing: 12) {
```
lucasnewman: Using a LazyVStack here switches to estimated layout sizes and ends up breaking the layout when generating large blocks of mark that switched to fully formatted once the streaming output is complete. 

There's more work to do to make that seamless, so I wouldn't make this change yet and keep the VStack to avoid breaking the chat experience. We pay a little bit for layout performance if the chat is long, but it's well within the scope of current devices to do it quickly enough that the performance tradeoff isn't worth the experience degradation.

lucasnewman:
@Blaizzy Overall this looks good and the modularization is very nice, see my comment though -- we shouldn't ship that change or it will be a regression.

PR #177, opened 2026-08-03
Dictation improvements

lucasnewman:
1) Use option + command for hands-free to avoid accidental activations when e.g. option-clicking
2) Allow dismissing dictation from the global overlay widget
3) Allow deleting individual transcripts or clearing all from the library
4) Show a "NEW" badge on the Extensions tab to allow for feature discovery for new and existing users, ask for mic permissions when enabling the dictation extension


Lazarus-931 (approved):
conflict, lgtm otherwise!

PR #178, opened 2026-08-03
Added Hugging Face LoRA adapter management

Yavaren:
## What changed

This PR adds LoRA adapter support for language models.

Adapters can now be found and downloaded from Hugging Face directly from the model configuration panel. Installed adapters can be selected, activated, paused during download, or deleted.

On the server side, adapters are validated before loading, including their configuration and compatibility with the selected base model. GGUF repositories are excluded.

I also added clearer download progress and error handling.

## Current scope

This supports adapters downloaded from Hugging Face only. Importing adapters from arbitrary local directories is intentionally not included in this PR.


Lazarus-931:
hi @Yavaren, thanks for the pr. i think this warrants a careful approach, do you have a visual of what this looks like?

and i think mlx-vlm downstream requires datasets>=2.19.1 as well for lora

Yavaren:
Hello @Lazarus-931! I attached all of the visuals. The adaptor selector is available through the model configuration panel or models page. 

I had a talk with Prince today regarding the looks and usability and got positive feedback. I am really looking forward to hear what you think!



Screenshot 2026-08-03 at 8 27 27 PM
Screenshot 2026-08-03 at 8 23 42 PM
Screenshot 2026-08-03 at 8 23 55 PM
Screenshot 2026-08-03 at 8 24 19 PM

Lazarus-931:
I think this is nice, lgtm!

merge conflicts btw

Yavaren:
@Lazarus-931 fixed the conflicts

Lazarus-931:
new conflicts!

PythonDistribution/Overlay/nativ_adapter_loader.py
```diff
@@ -0,0 +1,492 @@
...
+import argparse
+import json
+import math
+import os
+from dataclasses import asdict, dataclass
+from pathlib import Path
+from typing import Any, Mapping, Sequence
+
+
+LORA_SUFFIXES = (".lora_a", ".lora_b")
```
lucasnewman, line 12: Should this live in `mlx-vlm` instead, or is there a reason it has to be an overlay?

PythonDistribution/Requirements/mlx-vlm-server-macos-arm64.txt
```diff
@@ -30,7 +30,7 @@ miniaudio==1.71
 mlx-audio>=0.4.7
 mlx-lm==0.31.3
 mlx-metal>=0.32.0
-mlx-vlm>=0.6.8
+mlx-vlm==0.6.8
```
lucasnewman: We shouldn't pin specific revisions if possible, >= is actually what we want.

lucasnewman:
@Yavaren Looks cool! I had a couple of questions/comments inline but overall this seems good.

PythonDistribution/Overlay/nativ_adapter_loader.py
Yavaren: I agree we should add mlx-vlm long-term. Right now standard PEFT adapters are packaged differently, so their settings and weights need to be converted before mlx-vlm can use them. It also loads weights with strict=False so unmatched keys can be silently skipped.

PythonDistribution/Requirements/mlx-vlm-server-macos-arm64.txt
Yavaren: Agreed, my bad. Will fix right now.

PythonDistribution/Overlay/nativ_adapter_loader.py
lucasnewman: But the conversion is just remapping keys, right? I think this should just live in `mlx-vlm` since it's useful there too, and generally we're trying to keep the python overlay as minimal as possible here so we can consolidate effort for fixing bugs and maintenance of models in the server instead of split across two different surface areas. FWIW I don't think this should be a major lift, there's already some amount of LoRA support in `mlx-vlm` today that we can extend.

Yavaren: Yes, you are right. I had a call with Prince, he showed me something. Thanks a lot for pointing this out!

Yavaren: Lucas, do you have any specific plans for extending LoRA support in mlx-vlm?

lucasnewman: @Yavaren I didn't, but feel free to ping me if you want to chat about it or have a specific request.

PR #179, opened 2026-08-03
Sidebar: collapsible sections plus a collapse-all toggle

Lazarus-931:
The Pinned, Folders, and Sessions headers were static text, so the only thing that collapsed in the sidebar was an individual folder. Give each section a chevron matching the folder one, and add a single control that folds every section and folder at once.

Section state lives in NativSettings (three Bools, decoded with decodeIfPresent so existing installs default to expanded) and persists via the existing settings didSet. Per-folder state keeps using ChatFolder.isCollapsed; setAllFoldersCollapsed writes all folders with one saveFolders call.

The three headers now share one sidebarSectionHeader builder rather than repeating the chevron, with the Folders "new folder" button passed in as a trailing slot. Dropping a chat onto a collapsed Pinned or Sessions section expands it first, so the moved chat stays visible.

Collapse-all is reachable from the sidebar action bar and from a Collapse All Sections menu item, routed through ControlPanelNavigation's existing request-counter pattern.

looks like:

https://github.com/user-attachments/assets/01d48482-ce25-49e4-b597-0de873a9956d

### DIFF
--- a/Sources/Nativ/AppDelegate.swift
+++ b/Sources/Nativ/AppDelegate.swift
@@ -563,6 +563,11 @@ final class AppDelegate: NSObject, NSApplicationDelegate, NSMenuDelegate {
         showMainWindow()
     }
 
+    func toggleAllSidebarSections() {
+        controlPanelNavigation.collapseAllSections()
+        showMainWindow()
+    }
+
     func increaseChatFontSize() {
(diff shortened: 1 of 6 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #187, opened 2026-08-03
System Monitor: add die temperature, power, and fan sensors

johntdavies:
## Summary

The System Monitor covers CPU, GPU, memory and disk, but not the three signals that matter most when a Mac is running sustained local inference: **how hot the SoC is, how much power it is drawing, and whether the fans have spun up**. Those are also the axes that explain why decode throughput drifts over a long session — a model that looks slower an hour in is usually a thermal story, and right now Nativ's own analytics can't show that.

This adds a `SystemSensorSampler` that reads all three, plus a **Sensors** page to present them.

## Why this needs private interfaces

macOS exposes none of these through a supported API. `powermetrics` covers some of it but requires root, and **its `smc` sampler was removed in macOS 26**, so on Nativ's minimum OS it can no longer report die temperature at all. The three sources here are the ones the wider ecosystem (Stats, iStat Menus, mactop, macmon) settled on, and none of them need elevated privileges:

| Signal | Source |
|---|---|
| Temperature | `IOHIDEventSystem` sensor services, usage page `0xff00` |
| Power | IOReport `Energy Model` group, sampled as an energy delta |
| Fans | `AppleSMC` user client, `F<n>Ac` keys |

Every private symbol is resolved with `dlsym` at run time, and each reader is failable. A future OS that moves or drops one produces a **missing reading**, not a crash and not a link error. `ProcessInfo.thermalState` supplies thermal pressure and is ordinary public API.

## Design notes

**Power is layered, and labelled.** Rail coverage varies by chip. Where the CPU energy channels are populated the sampler reports a genuine SoC package total plus a CPU/GPU/ANE/DRAM breakdown. On an M5 Max, `GPU Energy` is the *only* energy channel in the entire IOReport tree — I enumerated all of them to confirm — so it falls back to whole-machine input power from the battery controller's `PowerTelemetryData`, and the UI states which of the two it is showing. Reporting a GPU-only figure as "SoC package power" would understate the machine by roughly two orders of magnitude, so the code requires a CPU rail before it will call a sum a package total.

**No sampling sleep.** A one-shot command line reader has to sleep ~1s to get an energy delta. The monitor already ticks at 1 Hz, so the IOReport subscription is held open and deltas are taken between successive collector passes. Cost per sample is microseconds.

**Handles kept open.** Sensor service lists, the SMC connection and the IOReport subscription are resolved once and reused, rather than re-enumerated every second.

**Chart de-duplication.** `SystemFPSHistoryChart` and the two new charts now share one `SystemValueHistoryChart` instead of repeating the same ~120-line Charts body three times, so this adds less UI code than the diff suggests.

## Verification

Built and run on an M5 Max (Mac17,6, macOS 26). Each reader was checked against an independent C implementation of the same interface:

| | This PR | Reference |
|---|---|---|
| Fans | `[2521, 2722]` RPM | `2521 / 2722` — exact |
| System input power | `49.31 W` | `SystemPowerIn=49313` mW — exact |
| Die temperature | `63.1 °C` (`PMU tdie6`) | 59–62 °C band |
| GPU power | `0.48 W` | 413–582 mW |
| Thermal pressure | `Nominal` → `Fair` under load | `powermetrics` agrees |

Two things worth flagging for review, both found during that verification:

- **`SystemPowerIn` is nested** inside the battery controller's `PowerTelemetryData` dictionary, not a top-level registry property. Tools that regex raw `ioreg` output miss this distinction.
- **`SMCKeyData` is 76 bytes in Swift but 80 in C.** Swift omits a *nested* struct's tail padding when embedding it; C includes it. The driver validates the struct size and rejects anything else. There is an explicit size guard so a layout mismatch fails closed rather than sending a malformed request — worth keeping if this is ever ported to another chip generation.

`MACOSX_DEPLOYMENT_TARGET` 14.0 is respected; no API above that baseline is used.

## Notes for the maintainer

- This is **private-API usage**. Fine for Sparkle/GitHub Releases distribution, but it would not pass App Store review, so it's worth a deliberate decision if that is ever on the roadmap.
- Fanless Macs correctly report "Fanless design" rather than an empty list; desktops without a battery controller simply omit system input power.
- I have not added a screenshot to `.github/screenshots/` — happy to add one in the house style if you'd like the Sensors page documented alongside the existing System Monitor shots.


Lazarus-931:
hi @johntdavies, this is awesome! we'd love this for system monitoring, do you have a visual on how this looks on top of main?

johntdavies:
Voila...

Screenshot 2026-08-04 at 01 23 26

johntdavies:
Just FYI, the code comes from my own NOC...
Screenshot 2026-08-04 at 01 33 07

Lazarus-931:
this is nice @johntdavies , one more, what does this look like on the status menu above?

lucasnewman:
@johntdavies This looks good to me, thanks! Would you mind fixing the conflict and then we can merge?

lucasnewman (approved):

PR #190, opened 2026-08-03
Dictation: fall back to on-device system speech recognition

johntdavies:
## Summary

Dictation needs an installed speech-to-text model **and** a running server. Until both are true, pressing the shortcut still records — and then hits a dead end, discarding the audio behind a *"download a speech model"* or *"start the server"* alert. That's the first thing a new user tries, and today it does nothing.

macOS ships speech models that need no weights of ours and no server. This uses them **when, and only when, the normal path cannot run** — first launch, a model still downloading, or a stopped server. It stays a fallback: the bundled MLX models are more accurate and cover far more languages. This just removes the dead end.

> Built on `SpeechAnalyzer` (macOS 26). The first commit used `SFSpeechRecognizer`; 9f38912 replaces it following @lucasnewman's review — see the thread. The old interface silently truncated anything past about a minute, which is covered under Verification below.

## What it does

- **`SpeechTranscriber`** does the transcription, with **`DictationTranscriber`** picking up the locales it doesn't cover — 45 vs 54, so the second tier is what gives Czech, Dutch, Polish, Russian, Swedish, Arabic and Hebrew speakers a fallback at all. `SpeechTranscriber` is the better recognizer where both apply: on a 97 s clip it kept sentence casing and punctuation that dictation lost.
- **Locale selection is `supportedLocale(equivalentTo:)`**, which resolves regions macOS has no model for onto ones it does — `en-JP` → `en-GB`, `pt-AO` → `pt-PT`, bare `de` → `de-DE`. A language neither module covers returns nil and the existing alert shows, rather than transcribing e.g. Welsh with an English model and inserting the result at the cursor.
- **Model assets.** macOS normally has the model already — system dictation draws on the same assets — and it only needs allocating to the app via `AssetInventory.reserve`, which downloads nothing. When a model genuinely is absent, the download runs in the background rather than holding the overlay open for an unknown length of time, and that recording gets *"macOS is downloading its \<language\> speech model — your recording is saved in Audio"* instead of an alert about a server the user may not have been trying to start.
- **The timeout scales with the audio length** rather than sitting at a flat 30 s, since recordings are no longer capped at a minute. A stalled recognizer still surfaces the normal alert instead of leaving the overlay spinning.

## Recognition is on-device by construction

`SpeechAnalyzer` has no server mode, so there is no code path in which audio leaves the Mac and nothing to opt out of. That also means no speech-recognition authorization, no permission prompt, and no `NSSpeechRecognitionUsageDescription` — `Configuration/Nativ-Info.plist` is untouched by this PR.

## Behaviour

The fallback reuses the existing success path, so its transcript is written to the `.txt` sidecar, recorded in analytics, and inserted at the cursor exactly like any other. It's attributed as `apple-speech (on-device)` in dictation history so it's distinguishable from an MLX transcript. When macOS can't help either, the original alert is shown unchanged — existing behaviour is preserved for anyone this doesn't cover.

## Verification

Tested on macOS 26 / M5 Max against speech synthesised with `say`, through the shipping code path:

| Input | Latency | Result |
|---|---|---|
| 5.6 s clip | 0.23 s | full transcript |
| 97.4 s clip | 1.29 s | full transcript, 1691 chars |
| 48 kHz stereo float | 0.13 s | full transcript |
| silence | — | `.empty` → normal no-speech feedback |

The 97 s clip is the one worth noting. `SFSpeechRecognizer` doesn't fail past its ~1-minute ceiling on a file — it truncates and reports success. The same recording came back as 659 characters, the *tail* of the audio, with the first minute silently gone and nothing to indicate it. That's the concrete cost of the original approach, and it's why the rewrite is a behaviour fix rather than an API tidy-up.

Three of the four bugs found while building the original `SFSpeechRecognizer` version were artefacts of that API and no longer exist: the authorization deadlock (nothing to authorize), the non-deterministic locale from an unordered `supportedLocales()` set (`supportedLocale(equivalentTo:)` is deterministic), and the released `SFSpeechRecognitionTask` (analysis is awaited, not escaping into a callback). The fourth — silence reported as an error — remains handled, as `.empty` surfacing the normal no-speech feedback.

## Notes

- Deliberately file-based rather than streaming, matching how dictation already records to disk before transcribing — it slots into the existing flow without changing audio capture. Live/partial transcription would be a much larger change and is not attempted here.
- No new dependency; `Speech` is a system framework.
- Happy to put this behind a settings toggle if you'd prefer it opt-in, or to prompt before any model download rather than starting it in the background. I left both automatic because they only activate where the app currently does nothing at all.

### DIFF
--- a/Configuration/Nativ-Info.plist
+++ b/Configuration/Nativ-Info.plist
@@ -30,6 +30,8 @@
 	<string>Copyright © 2025 Prince Canuma &amp; contributors</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Nativ records your microphone for dictation, voice notes, and meeting capture.</string>
+	<key>NSSpeechRecognitionUsageDescription</key>
+	<string>Nativ uses macOS on-device speech recognition to transcribe dictation before a speech model has been downloaded. Audio stays on your Mac.</string>
 	<key>NSAudioCaptureUsageDescription</key>
 	<string>Nativ records system audio when you start an audio capture.</string>
 	<key>NSScreenCaptureUsageDescription</key>
--- a/Configuration/Nativ-Info.plist
+++ b/Configuration/Nativ-Info.plist
@@ -30,8 +30,6 @@
 	<string>Copyright © 2025 Prince Canuma &amp; contributors</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>Nativ records your microphone for dictation, voice notes, and meeting capture.</string>
-	<key>NSSpeechRecognitionUsageDescription</key>
-	<string>Nativ uses macOS on-device speech recognition to transcribe dictation before a speech model has been downloaded. Audio stays on your Mac.</string>
 	<key>NSAudioCaptureUsageDescription</key>
 	<string>Nativ records system audio when you start an audio capture.</string>
 	<key>NSScreenCaptureUsageDescription</key>
--- a/Nativ.xcodeproj/project.pbxproj
+++ b/Nativ.xcodeproj/project.pbxproj
@@ -85,6 +85,7 @@
 		7936A52CFFF5AB09FC9CDDCD /* HuggingFaceCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = E9D1F1A2CDDD531089F289C9 /* HuggingFaceCache.swift */; };
 		7A86AD9E2E640430F47B94D5 /* FileTypeIcon.swift in Sources */ = {isa = PBXBuildFile; fileRef = 677258B04E9BB7CA8B89E11E /* FileTypeIcon.swift */; };
 		802B1F569D3D9FBE951EC8EA /* DashboardViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = ADEC2FFE7F1CFAC59170AC63 /* DashboardViewModel.swift */; };
+		826AB68248F3DD8952284745 /* AppleSpeechTranscriber.swift in Sources */ = {isa = PBXBuildFile; fileRef = 219898F64D1B7C28FA4A1B38 /* AppleSpeechTranscriber.swift */; };
 		86C75AF5CF63104E397339FB /* ArtifactsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = AD3387C33E274E567979F013 /* ArtifactsView.swift */; };
 		8B4DF732A8A6AA5FFBE11982 /* StatsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 64BBF79010B8120057237B52 /* StatsView.swift */; };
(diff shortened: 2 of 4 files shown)


lucasnewman:
@johntdavies Thanks for the contribution! What's the thinking behind using the older APIs vs the newer `SpeechAnalyzer`? The app specifically targets macOS 26+, so there shouldn't be a compatibility concern, and it doesn't have some of the onerous limitations like max 1 minute recording.

Lazarus-931:
> @johntdavies Thanks for the contribution! What's the thinking behind using the older APIs vs the newer `SpeechAnalyzer`? The app specifically targets macOS 26+, so there shouldn't be a compatibility concern, and it doesn't have some of the onerous limitations like max 1 minute recording.

from what i recall, don't we target macOS 14?


https://github.com/Blaizzy/nativ/blob/98c8b69942810516ed6ae22f08d0185b9c7f7c4d/project.yml#L10

lucasnewman:
> > @johntdavies Thanks for the contribution! What's the thinking behind using the older APIs vs the newer `SpeechAnalyzer`? The app specifically targets macOS 26+, so there shouldn't be a compatibility concern, and it doesn't have some of the onerous limitations like max 1 minute recording.
> 
> from what i recall, don't we target macOS 14?
> 
> https://github.com/Blaizzy/nativ/blob/98c8b69942810516ed6ae22f08d0185b9c7f7c4d/project.yml#L10

@Lazarus-931 The app target is 26+: https://github.com/Blaizzy/nativ/blob/98c8b69942810516ed6ae22f08d0185b9c7f7c4d/project.yml#L109

johntdavies:
@lucasnewman Apologies, my mistake. It came from a repo of mine where SFSpeechRecognizer is capped at 60 s per session. I'd written a handler around the error 209 restarts and stopped thinking about it. My iOS app still uses it, so it was reflex. I'd seen the macOS 14.0 in project.yml and forgotten the app target overrides it.

Rewritten on SpeechAnalyzer in c8500b6. It uses SpeechTranscriber, falling back to DictationTranscriber only for the locales that doesn't cover (Czech, Dutch, Polish, Russian, Swedish, Arabic, Hebrew and a few more).

Hopefully better now. :-)

lucasnewman:
@johntdavies Thanks! Can you fix the conflicts and then we can merge?

johntdavies:
Rebased onto `main` — it had drifted into conflict on `project.pbxproj` and `VoiceCaptureCoordinator.swift` (the shortcut-clash fix in #210 touched the same method). The pbxproj is a clean `xcodegen generate` rather than a hand-merge. Mergeable again.

Unrelated, but noticed while running the suite against the new base: `main` is red on `VoiceShortcutPreferencesTests.testDefaultsMatchExistingVoiceCommands`. It asserts `VoiceShortcut.recordDefault.displayName == "Fn + Control"`, but f9c870d changed the default to `[.control, .option, .command]` and the assertion wasn't updated. Nothing to do with this branch — both files are identical to `main` here. It goes unnoticed because `dev-build.yml` only runs `make xcode-build`; the tests aren't in CI.

Happy to send a one-line PR for it if useful.

PR #193, opened 2026-08-04
Mcp hub

Lazarus-931:
a wip pr to include mcp, tools and skills into existing extensions page

https://github.com/user-attachments/assets/b7de669a-6dab-4e9e-97fb-7499cdc06280

### DIFF
--- a/Sources/Nativ/Resources/MCPCatalog.json
+++ b/Sources/Nativ/Resources/MCPCatalog.json
@@ -14,7 +14,7 @@
     "name": "git",
     "summary": "Inspect and operate on Git repositories.",
     "command": "uvx",
-    "args": ["mcp-server-git"],
+    "args": ["--with", "mcp==1.12.0", "mcp-server-git"],
     "symbol": "arrow.triangle.branch",
     "tint": "orange",
     "sourceURL": "https://github.com/modelcontextprotocol/servers/tree/main/src/git"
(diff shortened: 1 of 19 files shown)


C-K-Loan:
@Lazarus-931 opened #206 against mcp-hub — adds the consent gate for MCP calls + wires up the shutdown that wasn't being called anywhere, plus a one-line fix for the test target not compiling. Should be a clean merge on top of what's here.

Sources/Nativ/Resources/MCPCatalog.json
```diff
@@ -0,0 +1,63 @@
...
+    "symbol": "brain.head.profile",
+    "tint": "purple",
+    "sourceURL": "https://github.com/modelcontextprotocol/servers/tree/main/src/memory"
+  },
+  {
+    "id": "sqlite",
+    "name": "sqlite",
+    "summary": "Query and edit a local SQLite database.",
+    "command": "uvx",
+    "args": ["--with", "mcp==1.12.0", "mcp-server-sqlite", "--db-path", "database.db"],
```
lucasnewman, line 58: Where does this file actually end up landing? I'm not sure what the working directory would be here.

lucasnewman (approved):
@Lazarus-931 Overall this looks reasonable. It does seem complex overall from an interface perspective, but I don't have any immediate ideas to reduce the complexity without playing around with it.

Sources/Nativ/Resources/MCPCatalog.json
Lazarus-931: changed it, it used to resolve to entire root system, now it's per server, by having MCPClient take an optional workingDirectory and set's process.currentDirectoryURL before launch, which should mean per-server dir like 
` ~/Library/Application Support/Nativ/MCP/<server-id>/` making stuff like sqlite be sort of sandboxed

Issue #191, opened 2026-08-03
Issues with voice


Lazarus-931:
1. When you try to replay voice recording, it doesn't have a pause or stop button, you have to wait for audio to finish
2. When I try to delete the audio while it's playing, UI looks like audio it's deleted, but in the background it continues playing
3. Sometimes when audio is pressed play, it routes to apple music

PR #197, opened 2026-08-04
Fix/voice playback 191

Lazarus-931:
Addresses the issues in #191, which exists becuase nativ didn't have it's own audio player, which is important for when you want to listen to the voice recordings, toggle them with UI and not have it launch external audio players(like apple music) for audio files.

Done by adding a AVAudioPlayer backend


lucasnewman (approved):
🙏

PR #154, opened 2026-07-30
MCP host support

Lazarus-931:
This adds a tools page that allows users to define mcps tools and use Nativ supported or community contributed ones.

With this, local models on nativ can use any mcp server users wish. 

This include instructions to how to contribute mcp server tools and a CI workflow that runs 24 hours to verify mcp servers return tool calls.

WIP

Screenshot 2026-07-31 at 9 49 39 AM

PR #141, opened 2026-07-28
Generalize the chat tool-call loop into a pluggable tool registry

C-K-Loan:
Hey Prince! Nativ's chat already has a solid multi-round tool-call loop, but it's currently hardwired to the built-in image tools. This PR generalizes that loop into a small pluggable registry so any tool can join, and adds four working native tools as examples — three read-only, one consent-gated write.

**What changed**
- `ChatToolRegistry`: a tool is now name + JSON schema + executor. The loop asks the registry for advertised schemas and dispatches incoming calls by dictionary lookup. Slim core, one file per tool under `Tools/`.
- The existing image tools are wrapped as registry entries; their code paths are unchanged.
- Three read-only native tools: `get_system_stats` (live data from the System monitor), `list_models` (installed MLX models), `get_server_stats` (request volume, tok/s, TTFT from the analytics store). Ask "how is my Mac doing" and the model answers with real numbers instead of guessing.
- One consent-gated write tool: `switch_model` lets the model change the loaded model on your explicit approval — a card renders in-chat with the requested model and an approve/deny choice; nothing happens until you decide.
- Expandable agent-step cells in the transcript: each tool round renders collapsed, one click shows the exact call, arguments, and raw result. Follows the existing reasoning-display pattern.
- One deliberate image-tool schema correction: the declared width/height minimum was raised from 64 to 256 to match the backend's actual floor, so schema-valid requests can no longer hit that backend rejection. Behavior is otherwise preserved.

**Zero regressions, verified**
- Image tool behavior compared against a clean-main baseline; the only intended difference is the declared minimum above.
- Full existing suite green, plus new unit tests for the registry, dispatcher, consent/cancel handling, and golden-pinned tool schemas.
- Rebased onto current main; no new dependencies.

**Demo**

The chained-reasoning centerpiece — one prompt, three tool calls in sequence, ending in a consent-gated model switch:

chained reasoning demo

Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/chained-demo.mp4

Individual tools:

`list_models`
list_models demo
Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/demo-list-models.mp4

`get_server_stats`
get_server_stats demo
Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/demo-get-server-stats.mp4

`get_system_stats`
get_system_stats demo
[...]

PR #200, opened 2026-08-04
Add MCP (Model Context Protocol) client support

C-K-Loan:
This adds MCP client support to Nativ's chat — any MCP server can plug into the same tool registry PR #141 introduced, alongside the native tools.

**What changed**
- A stdio JSON-RPC MCP client (`NativMCPClient.swift`), with `NativProcessController` extended to wire a child process's stdin (it previously only ever wired stdout/stderr for the app's own bundled server).
- Tool names are namespaced `mcp__<server>__<tool>` so servers can't collide with each other or with native tools.
- Servers are configured through a new **MCP Servers** settings panel — add/start/stop/remove, per-server tool-exemption list. **Nothing is bundled or hardcoded** — no server auto-starts or ships pre-configured; every server is explicitly added and explicitly started.
- Consent gating is fail-closed: any MCP tool call needs an explicit in-chat approve/deny, same pattern as `switch_model` from #141 — extended so a tool whose server can't be identified still requires approval rather than running quietly.
- 252/252 tests, zero regressions to existing native tools. Covered against a real fixture server, not a mock transport.

**On #154** — two layers of the same stack, not competing. This answers "can Nativ talk to MCP servers at all, safely." #154's catalog/CI-verification idea answers "how do we vet many community servers at scale" — additive, not conflicting.

**Demo**

`filesystem`
filesystem demo
Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/mcp-filesystem.mp4

`fetch`
fetch demo
Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/mcp-fetch.mp4

`memory`
memory demo
Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/mcp-memory.mp4

Multiple servers together in one conversation:
mcp ecosystem demo
Full quality: https://github.com/C-K-Loan/nativ/blob/demo-assets/demos/mcp-ecosystem.mp4

*(Browser/Playwright demo to follow shortly — server works, clip's still in review.)*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ DerivedData/
 __pycache__/
 .DS_Store
 AGENTS.md
+.playwright-mcp/
 /Configuration/Signing.local.xcconfig
--- a/Nativ.xcodeproj/project.pbxproj
+++ b/Nativ.xcodeproj/project.pbxproj
@@ -146,6 +146,7 @@
 		E4D299AADEA8470C37256DDE /* SystemMonitorStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3E540204DED30CFBDB0FE1AD /* SystemMonitorStore.swift */; };
 		E4D6CCEFDC2BA5A68FAE230A /* NativMCPClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8F5AE12A6EA1216750AFED5E /* NativMCPClient.swift */; };
 		E64F3E5C624F0C949AE70C60 /* SoftwareUpdater.swift in Sources */ = {isa = PBXBuildFile; fileRef = 822185ADAAC5DA3F034179B0 /* SoftwareUpdater.swift */; };
+		E6C6A47E0949A0087BCE08D8 /* MCPServersView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0DCD43578DBAFBEC2F0D651D /* MCPServersView.swift */; };
 		E96652149FDCCF65899280A8 /* IntegrationsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 41417C210BF817681A8C6D87 /* IntegrationsView.swift */; };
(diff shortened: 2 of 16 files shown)


C-K-Loan:
@Lazarus-931 — noticed #193 after opening this one. Looks like we ended up solving the same problem independently, quick comparison of our branches:

- **Protocol layer:** #193 wraps the official mcp-swift-sdk; this PR hand-rolls the stdio/JSON-RPC transport. Less protocol code to maintain long-term is a genuine advantage for #193's approach here — worth adopting regardless of which core ends up landing. This PR's upside is the flip side: zero external dependency for MCP support, no version pin or supply-chain surface to track — real tradeoff either way.
- **Catalog & discovery:** #193's server catalog plus the verify-mcp-catalog.yml CI check is a real gap this PR doesn't address at all. Vetting community servers at scale is exactly the kind of infra this repo needs, independent of whichever client sits underneath it.
- **Consent gating:** this PR routes every MCP tool call through the same approve/decline gate as switch_model (#141), fail-closed even when a server can't be identified — covered by dedicated tests for that boundary. From reading #193's current state, MCP calls execute without that gate yet. Could well just be sequencing on a WIP, but flagging it since it's worth locking down early.
- **Test coverage:** 266/266 here, against a real subprocess fixture rather than mocks.

Given the split — #193 ahead on protocol choice and discovery UX, this one ahead on the safety/execution path — there's probably a stronger combined version than either alone. Happy to help wire the consent-gating and process-lifecycle handling into #193's catalog/SDK approach if that's useful, or split it however @Blaizzy prefers.

Lazarus-931:
hi @C-K-Loan, yeah so I'd say closing this an leaving comments in #193 if what else to include on it, happy to merge with u as contributor!

C-K-Loan:
Thanks sounds good sending a PR to your branch!

C-K-Loan:
Closing in favor of #193 — the consent-gating + lifecycle pieces went over as #207/#208/#209.

PR #201, opened 2026-08-04
Make user prompts editable and copyable

Yavaren:
## Summary

- make user prompts reliably selectable and add a copy action consistent with assistant responses
- allow earlier user prompts to be edited in the existing composer and regenerated
- preserve any unsent draft and attachments while editing, restoring them after save or cancellation
- disable editing while the session is busy or the current model configuration is unavailable

Resending an edited prompt updates that prompt in place, removes all later conversation items, persists the revised transcript, and generates a fresh assistant response using the current model settings.


Yavaren:
Visuals of the UI changes. 

Now shows copy and edit buttons when hovered over the user prompt 

image

After clicking the edit button, you get to see this view, where you can cancel editing and send the updated prompt

image

Lazarus-931 (approved):
lgtm!

PR #203, opened 2026-08-04
support routine

Lazarus-931:
wip of supporing routines locally

https://github.com/user-attachments/assets/42ffa553-cd21-4850-9058-a1583a0ffca4


Sources/Nativ/ControlPanelView.swift
```diff
@@ -12,6 +12,7 @@ enum ControlPanelTab: String, CaseIterable, Identifiable {
     case models = "Models"
     case integrations = "Integrations"
     case extensions = "Extensions"
+    case routines = "Routines"
```
lucasnewman, line 15: This is ok for now but this list can't grow indefinitely. I wonder if a better design would be allowing any chat session to be scheduled as a routine, and only then the item shows up in the listing for progressive disclosure of complexity?

lucasnewman (approved):

Sources/Nativ/ControlPanelView.swift
Lazarus-931: yeah, tried playing with that idea, and I actually think is better.

we can have chat be the place to schedule routines, either on the same chat or a new one, then the Routine page only appears when a routine runs for the first time.

i'd disappear if there is no routines scheduled

I did see the navbar become bloated

Lazarus-931:
another improved pr forthcoming

Sources/Nativ/ControlPanelView.swift
Lazarus-931: coming back to this, i think we should have routine feature for new or existing chats and mark chats with routines set to run on them(handling them being deleted and all that)

then we can remove this "Routine" page

Lazarus-931:
see #224

PR #206, opened 2026-08-04
Add consent gating + process lifecycle to mcp-hub

C-K-Loan:
Adds the two things flagged on #200: a consent gate for MCP tool calls (fail-closed, including when a server can't be identified) and wiring `MCPHostManager.shutdown()` into app termination, since it existed but nothing called it — both covered with tests against real subprocesses. Also fixes the test target, which wasn't compiling (missing NativSkill.swift), so the existing suite was being silently skipped. Nothing else touched.


Lazarus-931:
sweet, could u make this a stacked pr, easier to handle conflicts

C-K-Loan:
Split into a stack per your suggestion: #207 (test fix) → #208 (consent gate) → #209 (lifecycle). Closing this one.

PR #207, opened 2026-08-04
Fix NativTests target: add missing NativSkill.swift

C-K-Loan:
Test target wasn't compiling — NativSettings.swift references NativSkill but the file was never added to the target, so the whole suite was being silently skipped instead of run. One line.


lucasnewman (approved):
Thanks!

PR #208, opened 2026-08-04
Require consent before running an MCP tool

C-K-Loan:
Stacked on #207 (includes its commit too — GitHub can't base a fork PR on another fork-only branch, so this shows combined until #207 merges). The new commit here is bbb0770: routes MCP calls through the same consent gate switch_model already uses, fail-closed even when a server can't be identified. Tested.


Lazarus-931 (approved):
👍🏽

PR #211, opened 2026-08-05
Fix dictation shortcut clashing with macOS navigation

### LINKED ISSUE #210: Default global dictation shortcut clashes with FN + arrows for Spaces
Spaces uses FN + Arrows to switch. The default diction shortcut is FN + Ctrl. So the dictation pops up every time you switch spaces.

I see you can assign an override but you should offer an option that turns off global shortcut completely. Also should pick a default that will not get triggered by a standard macOS shortcut.

Lazarus-931:
Closes Blaizzy/nativ#210.

The default global dictation shortcut was `Fn + Control`, a modifier-only chord matched on an exact modifier comparison. Because those modifiers are held during Spaces / Mission Control navigation, dictation fired every time the user switched spaces.

## Changes

* **Non-clashing default:** the default dictation shortcut is now **Control + Option + Command** instead of `Fn + Control` — a chord macOS binds to nothing and that isn't held during navigation.
* **Double-tap to start (hands-free):** a modifier-only shortcut now requires a **double-tap within 0.4s** to start hands-free dictation, so an accidental single chord press no longer triggers it. Push-to-talk (hands-free off) still holds-to-talk; keyed shortcuts are unchanged.
* Updated the now-stale "No Recent Recording" hint to use the live shortcut names, and the dictation extension's `defaultShortcut` label.

## Notes

* Per review feedback, the standalone "Enable dictation shortcut" toggle was dropped: the extension's own enable/disable state is the single switch, since the shortcut is currently the only way to trigger dictation. A dictation button in the chat composer (which would justify an independent toggle) is a good follow-up.
* Existing preferences are preserved; the new default only applies to fresh installs / restored defaults.
* `VoiceModifierToggleShortcutState` tests were rewritten for the double-tap behavior with an injectable clock.

### DIFF
--- a/Sources/Nativ/Features/VoiceCapture/AudioView.swift
+++ b/Sources/Nativ/Features/VoiceCapture/AudioView.swift
@@ -1669,6 +1669,15 @@ struct AudioView: View {
         }
     }
 
+    private var handsFreeDescription: String {
+        guard shortcuts.isHandsFreeEnabled else {
+            return "Hold while speaking; release to transcribe."
+        }
+        return shortcuts.recordShortcut.keyCode == nil
+            ? "Double-tap to start; double-tap again to transcribe."
+            : "Press once to start; press again to transcribe."
+    }
+
     private var shortcutConfigurationPanel: some View {
         VStack(alignment: .leading, spacing: 18) {
             VStack(alignment: .leading, spacing: 3) {
@@ -1684,13 +1693,9 @@ struct AudioView: View {
                     VStack(alignment: .leading, spacing: 3) {
                         Text("Hands-free")
                             .font(.callout.weight(.medium))
-                        Text(
-                            shortcuts.isHandsFreeEnabled
-                                ? "Press once to start; press again to transcribe."
-                                : "Hold while speaking; release to transcribe."
-                        )
-                        .font(.caption)
-                        .foregroundStyle(.secondary)
+                        Text(handsFreeDescription)
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
                     }
 
                     Spacer(minLength: 16)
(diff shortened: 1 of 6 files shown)


Sources/Nativ/Features/VoiceCapture/AudioView.swift
```diff
@@ -1682,20 +1691,20 @@ struct AudioView: View {
             VStack(alignment: .leading, spacing: 10) {
                 HStack(spacing: 12) {
                     VStack(alignment: .leading, spacing: 3) {
-                        Text("Hands-free")
+                        Text("Global dictation shortcut")
```
lucasnewman: It's not possible to trigger dictation at all if this is off, right? Why wouldn't this be gated behind the dictation extension being enabled? I'm trying to imagine a scenario where it makes sense to have the extension enabled but the shortcut off and can't think of one, but I might be missing something.

ascarter: ChatGPT/Codex I think put a voice icon in the chat window to trigger dictation on demand. That's a feature ask I guess but if you had that, then it would make sense to disable the shortcut independently of disabling dictation support overall. But given how Nativ works at the moment, agree that turning it off makes sense since you can't use it without a shortcut.

lucasnewman: Yeah, I think that's a good idea, having dictation as a button in the chat composer could work well.

(Also, hi Andrew! It's been a while since our last Seattle Xcoders meeting -- hope you're doing well!)

lucasnewman (approved):

Sources/Nativ/Features/VoiceCapture/AudioView.swift
ascarter: Thanks! Excited about this project. Hope to see you back in Seattle sometime!

PR #214, opened 2026-08-05
Calibrate the param size estimate for image models

Lazarus-931:
Calibrating the param estimate for image models.

Prior, the advertised model size was taken straight from HuggingFace's `safetensors` parameter summary — `Σ params[dtype] × bits/8`. For image-generation (diffusion) repos that summary only counts the diffusion **transformer** and omits the text encoder(s), VAE, and other components that actually get downloaded, so the size shown sat well **below** the real download:

| Model | Shown (param estimate) | Real files |
|---|---|---|
| FLUX.1-dev | ~24 GB | ~58 GB |
| Stable Diffusion 3.5 Large | ~16 GB | ~72 GB |
| Qwen-Image | ~41 GB | ~58 GB |

(And repos with no `safetensors` summary showed no size at all, then downloaded ~24 GB.)

This scales the parameter estimate up for image-generation / image-editing models so the advertised size — and the disk-capacity check and download-progress denominator that reuse it — land closer to the real download. The memory estimate stays on the raw parameter figure (that reflects the transformer in memory).

Minimal by design: it's a calibration, not exact. The precise fix is to sum the repo's real file sizes via the HF `?blobs=true` API for the selected model, which also covers the no-summary case — left as a follow-up.

### DIFF
--- a/Sources/Nativ/Features/Models/HuggingFaceHub.swift
+++ b/Sources/Nativ/Features/Models/HuggingFaceHub.swift
@@ -99,18 +99,42 @@ struct HuggingFaceModel: Decodable, Identifiable, Equatable, Sendable {
         }
 
         provider = LocalModelProviderResolver.resolve(repoID: id, modelType: nil, architectures: [])
-        sizeBytes = safetensors?.sizeBytes
         capabilities = Self.resolveCapabilities(
             pipelineTag: pipelineTag,
             libraryName: libraryName,
             tags: tags
         )
+        let parameterBytes = safetensors?.sizeBytes
+        sizeBytes = Self.calibratedDownloadBytes(parameterBytes, capabilities: capabilities)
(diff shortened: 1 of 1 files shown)


lucasnewman (approved):
Ok, good enough without making network calls I suppose :) Should we just do an async fetch to get the real size once we've identified it as an image model, though?

Lazarus-931:
> Ok, good enough without making network calls I suppose :) Should we just do an async fetch to get the real size once we've identified it as an image model, though?

yes actually I will open a new pr instead, the issue was those async calls for model loaded on screen, but I think we can cache those for max 2 calls per model on screen or something like that

Lazarus-931:
Superseded by #216 — an accurate async, cached fetch of the model's real file sizes instead of a fixed multiplier.

Issue #131, opened 2026-07-27
Guide chat image generation when no image model is selected


Blaizzy:
## Problem

Chat image generation silently becomes unavailable when no image-generation model is activated in the Models tab.

The current Chat request builder only advertises `generate_image` and `edit_image` when `imageGenerationModelID` is already set. If it is unset, the language model receives no image tools and cannot explain what is missing, list installed image models, or help the user select one.

This is especially confusing because the Images tab can fall back to an installed/default image model while Chat still has no configured model.

## Expected behavior

A tool-capable chat model should guide the user through image-model selection before attempting generation.

1. When the user asks to generate an image, determine whether an image-generation model is currently selected or loaded.
2. If no model is selected but compatible image-generation models are installed:
   - list the installed compatible models by display name and repository/model ID;
   - ask the user which one to use;
   - do not start generation until the user selects one.
3. After selection, retain the selection for the chat request/session and invoke image generation with the exact selected model ID.
4. The image-generation request must explicitly pass the selected model name/ID rather than relying on an implicit global default.
5. If no compatible image-generation models are installed, tell the user to download one from the Models tab and do not attempt generation.
6. If editing an image, only offer models that support the required image-editing capability.

## Suggested flow

```text
User: Generate an image of a mountain lake.

Assistant: No image-generation model is selected. Installed models:
1. flux2-klein-4b-8bit — mlx-community/flux2-klein-4b-8bit
2. FLUX.2-klein-4B — black-forest-labs/FLUX.2-klein-4B

Which model should I use?

User: Use the first one.

Assistant tool call:
generate_image(
  model: "mlx-community/flux2-klein-4b-8bit",
  prompt: "A mountain lake..."
)
```

If none are installed:

```text
No compatible image-generation model is installed. Download one from the Models tab, then try again.
```

## Implementation notes

- Do not hide every image-related capability solely because `settings.imageGenerationModelID` is `nil`.
- Give Chat access to the locally discovered image-generation models and their capabilities.
- Add an explicit `model`/`model_id` argument to the image-generation tool contract, or otherwise make the selected model ID part of the tool execution request.
- Validate that the requested model is installed and supports generation/editing before execution.
- Define whether the chat choice is session-scoped or also updates the global preload selection; avoid silently changing global preload state without a deliberate product decision.
- The Images tab and Chat should use the same model resolution rules.

## Acceptance criteria

- [ ] With a preselected image model, Chat generates using that exact model.
[...]

PR #215, opened 2026-08-05
LLM image generation tool calling feature request #131

Yavaren:
Summary
Adds image generation and editing tool calling to chat.
When the LLM requests an image operation, the app checks the models installed locally and filters them by the required capability. If there isn’t already a compatible model selected for the chat, the user sees a native model picker. The selected model is validated and remembered for future image requests in that chat.
Model selection is handled entirely by the app—the LLM never provides or controls model identifiers. This prevents invalid, escaped, or hallucinated model IDs from breaking image generation.
If no compatible image model is installed, the app explains that one needs to be downloaded from the Models page.


Yavaren:
Here are some UI components of the update by each state

No image model installed
Screenshot 2026-08-05 at 10 15 09 PM


LLM is asking you to select one installed image model
Screenshot 2026-08-05 at 10 14 18 PM


Image generation is completed
Screenshot 2026-08-05 at 10 14 28 PM


Image editing (runs on a previously selected image model)
image

lucasnewman (approved):
LGTM, thanks!

PR #216, opened 2026-08-05
fetch model size using async calls, cached

Lazarus-931:
instead of doing local calculations like in #214, we'd instead make async calls to hf for model size, addresses the issue of inaccurate model param report


Lazarus-931:
@lucasnewman,

so i closed #214 and instead made it a backup incase the async doesn't work(d38256d)

async does like under 200ms and was near instant

https://github.com/user-attachments/assets/3936dcd2-acee-4546-81be-ca842ce4476c

lucasnewman (approved):
Nice!

PR #217, opened 2026-08-05
Fix hub discover pagination: filter at the API + paginate over visibl…

Lazarus-931:
The discover search fetched top-sorted safetensors models capability-agnostic and only filtered by capability/access client-side after slicing a raw page, so pages showed a varying count and Trending/Downloads for image models could come back empty. Map a single selected capability to a HuggingFace pipeline_tag so the API returns matching models, re-fetch when the capability/access filters change, and paginate over the filtered (visible) results so every page holds a consistent count.

Discovered while doing #216, and below is what it used to look like: 

https://github.com/user-attachments/assets/be0721a3-e142-4474-9ba8-41daec08f74f


lucasnewman (approved):
Good catch!

PR #218, opened 2026-08-06
permission setup during onboarding

Lazarus-931:
WIP of setting up all setting permissions for users during the onboarding phase.

https://github.com/user-attachments/assets/7b4109d3-c22a-447e-9a57-d53e29a2ad46

### DIFF
--- a/Sources/Nativ/Features/Settings/NativPermissionsCard.swift
+++ b/Sources/Nativ/Features/Settings/NativPermissionsCard.swift
@@ -106,11 +106,11 @@ struct NativPermissionsSummary: View {
         let outstanding = store.outstandingPermissions.count
         switch outstanding {
         case 0:
-            return "All set. Nativ has everything it needs."
+            return "All voice permissions are on."
         case 1:
-            return "1 permission still needed."
+            return "1 optional permission is off."
         default:
-            return "\(outstanding) permissions still needed."
+            return "\(outstanding) optional permissions are off."
         }
     }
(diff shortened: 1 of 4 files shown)


lucasnewman:
I think we should soften the language here from "needed" to "optional", as these permissions could been seen as pretty invasive (it allows the app to record and control everything on your device), and we don't actually need them unless you enable the voice dictation feature. The app is perfectly useful without them enabled.

lucasnewman:
@Lazarus-931 Also, where is the "Add a voice" feature coming from in your video? Is that just a local change? We definitely don't want to use Kokoro by default...

Lazarus-931:
> @Lazarus-931 Also, where is the "Add a voice" feature coming from in your video? Is that just a local change? We definitely don't want to use Kokoro by default...

yeah it's a local dev version

Lazarus-931:
Screenshot 2026-08-06 at 1 30 15 PM

Blaizzy:
Why do we need screen recording ?

Blaizzy:
Why do we need voice and kokoro?

Lazarus-931:
> Why do we need screen recording ?


Screenshot 2026-08-06 at 1 48 12 PM

Lazarus-931:
> Why do we need voice and kokoro?

voice permission? for audio?

kokoro is local dev, ignore

lucasnewman (approved):
🙏

PR #221, opened 2026-08-06
Fix user prompt layout breaking when clicked

Yavaren:
## Summary

Fixes a layout issue introduced with selectable user prompts in #201.

Clicking a user prompt could cause its text to become clipped inside the message bubble. The selectable prompt now uses a stable text view that keeps the same layout when focused while preserving text selection, copying, wrapping, and font scaling.


lucasnewman (approved):
LGTM

Blaizzy:
This PR tests are failing

PR #222, opened 2026-08-06
Sidebar UI improvements

lucasnewman:
1) Improve disclosure arrow styling and layout
2) Hide Pinned and Folders when they are empty so we can keep it clean. Entering selection mode will show the empty groups, and pinning a session will automatically make it visible, etc
3) Fix an animation block applying to the whole control panel and making quick clicks between views fail to register

Screenshot 2026-08-06 at 11 14 19 AM
Screenshot 2026-08-06 at 11 14 32 AM


Lazarus-931:
yes, this was actually my doing, thanks :)

Lazarus-931 (approved):
👍🏽

PR #224, opened 2026-08-07
Feature/routines chat first

Lazarus-931:
supersedes #203, where "Routines" page is removed and the UI follows a much simpler approach.

Users create recurrent routines from exiting or same button used by chat creation, and this marks chats with a sign to symbolize recurrent routine.

https://github.com/user-attachments/assets/f288acb7-f040-437b-8d81-465ba9df8225

Users can't delete a chat without approving it's a recurrent job as well:

Screenshot 2026-08-07 at 8 08 30 AM


lucasnewman (approved):
Looks good!

PR #225, opened 2026-08-07
Check to make sure all shards are downloaded for sharded VLM/LM models

lucasnewman:
Inspired by https://github.com/Blaizzy/nativ/pull/8 but a little bit of a different approach -- use the shard manifest to make sure all shards are present before declaring the model downloaded and usable. The project file was also out of date so I regenerated it as well.


Lazarus-931 (approved):
lgtm!

PR #227, opened 2026-08-08
Fold image generation into chat; remove the standalone Images tab

Lazarus-931:
long over do pr for merging image gen into chat

I think this frees up navbar and seems more natural to users who never used a sep chat and image feature in other model platforms.


lucasnewman:
This means users won't be able to use image models standalone without a language model that supports tool calling, fwiw, which is why the image view exists, but go for it if you thinks that's reasonable / discoverable.

lucasnewman (approved):

Blaizzy (changes requested):
I agree with Lucas here.

The idea is not to remove the image generation page but rather place it in within the chat window as a tab, the same way you have chat and cowork for example.

Image

PR #228, opened 2026-08-08
nativ cli

Lazarus-931:
**Commands**
- `run` — one-shot completion (streams; `--system`, `--image`, `--json`, stdin)
- `chat` — interactive REPL (`/model`, `/system`, `/reset`, `"""` multiline)
- `serve` / `stop` / `status` — manage the bundled server (`serve -d` waits until ready)
- `models list | use | pull | rm` — list loaded + locally-cached models, set default, download, remove
- `embed` — text embeddings
- `image` — image generation
- `transcribe` — speech-to-text
- `audio speak` — text-to-speech (the `audio` group grows into separation/enhancement/sfx as the engine exposes those endpoints)
- `config show | set | path` — the app writes `cli.json`; this configures the CLI standalone
- `agent` — a full reference for coding agents (markdown or `--json`)

**Design**
- **Contract = the engine, not new logic.** Inference goes through the server's OpenAI-compatible API; the CLI carries only a thin HTTP client, no reimplementation of engine behavior. Per-capability models (chat/embedding/image/stt/tts) resolve flag → env → `cli.json` → default.
- **Decoupled.** The only app touchpoint is `cli.json`, so Nativ internals can change without breaking the CLI, and the same binary works against a local *or* remote server.
- **Tested.** Unit tests cover SSE parsing, multipart framing, and config resolution; verified end-to-end against a real mlx-vlm server (chat/embeddings/models/config all green).

**Direction (follow-ups)**
Make the inference commands server-aware with a local fallback — reuse the running server when present, otherwise cold-load via the bundled engine CLI — so the CLI works whether or not the app is open. Offline tooling (`convert`) will shell straight to the engine rather than being reimplemented.

---

*Note on the `agent` command:* it exists so a coding agent can learn the whole CLI in one cheap call instead of probing `--help` or the repo. Quick measurement — two fresh agents asked to produce a full usage guide, one told about `nativ agent`, one left to figure it out:

| | Without `agent` | With `agent` |
|---|---|---|
| Tokens | ~38,400 | ~18,600 |
| Tool calls | 9 | 1 |
| Wall time | 172 s | 48 s |
| Result | reached a wrong conclusion | correct + complete |

≈2× fewer tokens, one call instead of nine, and it avoided a mistaken finding. (n=1 per side — a broader multi-model run is planned.)


Blaizzy:
> ~~transcribe~~
> audio ~~speak~~

The resulting api arg should be:

`audio --task [tts, stt, sts, vad, lid] `

Where the tasks are the same as in mlx-audio.

Blaizzy:
Regarding the agent, thanks for sharing the results, it does indeed provide great time savings!

Could you share the inputs and traces as well?

Lazarus-931:
Reran as a clean A/B — the difference is baked into the **binary**, not the prompt.

## Reproduction

Two builds of the **same commit**; in one, the `agent` command is removed entirely (no trace of it in `--help`). Two fresh agents (independent contexts, same model) each get the **identical** prompt — no hint about `agent`, no restriction:

> Learn how to use the `nativ` CLI (binary: `<binary>`), then write a short usage guide covering all its commands.

So the only variable is whether `agent` exists and is self-discoverable via `--help`. Token usage is measured by the runtime.

## Results

| | no `agent` | `agent` present (no hint) |
|---|---:|---:|
| Tokens | 44,623 | 26,612 |
| Tool calls | 13 | 8 |
| Wall time | 140 s | 102 s |

With `agent` in the binary — found on its own via `--help`, never hinted — the agent used **~40% fewer tokens** and fewer tool calls. Without it, the agent fell back to reading `--help` across every subcommand and then the source files.

n=1 per side (agent behavior is stochastic), but the direction is consistent across runs: a self-documenting `agent` command is a cheap entry point for coding agents, with no downside when it isn't used.

Blaizzy:
The tests fail

Also, have you updated to wrap the mlx-vlm CLI tools as we spoke or this is more of temporary solution till we build it in mlx-vlm?

Lazarus-931:
> The tests fail
> 
> Also, have you updated to wrap the mlx-vlm CLI tools as we spoke or this is more of temporary solution till we build it in mlx-vlm?

this is a temp solution, but i think i can push upstream pr so i'll do that before changing this to full supported wrap

PR #229, opened 2026-08-08
chat title shimmer

Lazarus-931:
Until model output is done, chat name in navbar shimmers to signal model is generating

https://github.com/user-attachments/assets/067be0e7-3eb8-4db7-8210-01bc0abcae5b

also some glitches with renaming fix

### DIFF
--- a/Sources/Nativ/ControlPanelView.swift
+++ b/Sources/Nativ/ControlPanelView.swift
@@ -3819,6 +3819,11 @@ private struct ControlPanelRecentSessionRow: View {
                         .onExitCommand {
                             isRenaming = false
                         }
+                        // Clicking away ends the rename (commit) instead of
+                        // leaving a stuck field/caret that swallows clicks.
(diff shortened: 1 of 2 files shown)


lucasnewman:
The shimmer in the sidebar seems a little bit distracting? I think it looks good in the "Working" thinking area though. Maybe if it's slowed down and a bit subtler?

Lazarus-931:
> The shimmer in the sidebar seems a little bit distracting? I think it looks good in the "Working" thinking area though. Maybe if it's slowed down and a bit subtler?

made it softer, closer to how the collapsible reasoning/"thinking" panel inside the chat transcript was

lucasnewman:
Can you check to make sure this isn't increasing baseline cpu usage? Otherwise seems ok to me.

Lazarus-931:
> Can you check to make sure this isn't increasing baseline cpu usage? Otherwise seems ok to me.

nope, title shimmer does not increase baseline CPU

Issue #226, opened 2026-08-07
Server startup fails entirely when the preloaded speech-to-text model can't load — should degrade gracefully


rymalia:
> *This was generated by AI during triage.*

## Summary

If the selected speech-to-text model fails to load during server startup, the entire `mlx-vlm-server` process exits (status 3) — even though the language model had already loaded successfully seconds earlier. The app then appears completely broken: no chat, no models, and the server crash-loops on every restart until the user figures out that the *speech* model is the culprit and deselects it.

A missing optional capability (dictation) shouldn't kneecap the whole application. Startup should degrade gracefully: log the audio-model failure, surface it in the UI, and continue serving the language model.

## Environment

- Nativ 0.2.2, installed from the latest release DMG
- macOS 26, Apple silicon
- Speech models tried: `aufklarer/Qwen3-ASR-0.6B-MLX-4bit` and `aufklarer/Qwen3-ASR-1.7B-MLX-8bit` (both fail the same way)
- Language models tried: `Qwen/Qwen3-1.7B`, `LiquidAI/LFM2.5-1.2B-Instruct-MLX-8bit`, `mistralai/Ministral-3-3B-Instruct-2512` — **all load successfully** before the crash

## Steps to reproduce

1. In Nativ, select a language model and one of the `aufklarer/Qwen3-ASR-*-MLX-*` speech-to-text models for preload.
2. Start the server (or launch the app).
3. The language model preloads fine, then the speech-model preload throws, FastAPI's lifespan aborts, and the server exits with status 3. Repeat forever.

## Log excerpt

```
2026-08-07 16:34:57,050 - INFO - Model and processor loaded successfully.
2026-08-07 16:34:57,050 - INFO - Language model ready, continuous batching enabled.
2026-08-07 16:34:57,050 - INFO - Pre-loading speech-to-text model: aufklarer/Qwen3-ASR-1.7B-MLX-8bit
...
OSError: Can't load feature extractor for '/Users/.../models--aufklarer--Qwen3-ASR-1.7B-MLX-8bit/snapshots/e5450a2...'.
... make sure '<snapshot path>' is the correct path to a directory containing a preprocessor_config.json file

fastapi.exceptions.HTTPException: 500: Failed to load audio model: Can't load feature extractor for ...
ERROR:    Application startup failed. Exiting.

mlx-vlm-server stopped with status 3
```

[...]

PR #231, opened 2026-08-08
Skip pre-loading a speech model that cannot load

Lazarus-931:
When a selected speech-to-text model can't load, pre-loading it aborts server startup and takes chat down with it (#226); this detects that case before launch and starts the server without --stt-model, logging which model was skipped and why, so a broken dictation model can no longer brick the app, complementing the upstream fix in https://github.com/Blaizzy/mlx-vlm/pull/1814, which this does not depend on.


lucasnewman (approved):
LGTM

Issue #230, opened 2026-08-08
Voice dictation shortcuts are unreliable in hands-free and push-to-talk modes


Blaizzy:
## Summary

Voice dictation activation is unreliable in both **Hands-free** and **push-to-talk** modes. This is a combination of shortcut migration inconsistencies, a timing-sensitive modifier gesture, and 
permission recovery gaps.

### 1. The shortcut migration is incomplete

Commit [`f9c870d`](https://github.com/Blaizzy/nativ/commit/f9c870d3a446ede7f6b572fa1eebda58e2d31f46) changed the runtime default from **Fn + Control** to **Control + Option + Command** because Fn + Control conflicted with macOS navigation.

However:

- Existing valid preferences are always restored and are never migrated from the old default in [`VoiceShortcutPreferences.init`](https://github.com/Blaizzy/nativ/blob/14761a1/Sources/Nativ/Features/VoiceCapture/VoiceShortcut.swift#L163-L181).
- Affected existing installations therefore remain on **Fn + Control**.
- [The bundled extension manifest](https://github.com/Blaizzy/nativ/blob/14761a1/Extensions/VoiceDictation/Manifest.json#L31-L40) still advertises Fn + Control.
- [The runtime fallback manifest](https://github.com/Blaizzy/nativ/blob/14761a1/Sources/Nativ/Features/Extensions/VoiceDictationExtension.swift#L123-L133) advertises Control + Option + Command.
- [The unit test](https://github.com/Blaizzy/nativ/blob/14761a1/Tests/NativTests/AudioTests.swift#L657-L669) still expects Fn + Control while production returns Control + Option + Command.

### 2. Modifier-only hands-free activation is timing-sensitive

Modifier-only hands-free activation now requires two clean taps within only `0.4` seconds in [`VoiceModifierToggleShortcutState`](https://github.com/Blaizzy/nativ/blob/14761a1/Sources/Nativ/Features/VoiceCapture/FnControlShortcutMonitor.swift#L29-L88).

Any regular key press marks the modifier gesture as a chord and invalidates it. This is especially difficult with a two- or three-modifier shortcut.

Push-to-talk uses an exact modifier comparison in the same monitor. A momentary missing or additional modifier flag causes the held state to end, making recording stop unexpectedly.

### 3. Development builds can lose global-input permissions

The shortcut monitor uses `NSEvent.addGlobalMonitorForEvents`. Apple documents that key-related global events require Accessibility trust:

[...]

Issue #163, opened 2026-07-31
Extension platform: declarative dashboards and typed multimodal workflows


Blaizzy:
## Summary

Extend Nativ's extension platform so external extensions can define:

1. **Typed, multimodal workflows** composed from allowlisted host operations.
2. **Declarative custom dashboards** rendered from approved native components.

This should let extensions add meaningful behavior and UI without adding extension-specific Swift code to `nativ-core` or executing arbitrary code inside the host.

## Motivation

The current declarative workflow runtime is optimized for one text transformation:

```text
Read selected text → Run language model → Replace selection
```

That works for Rewrite-style extensions, but it is not sufficient for:

- Vision workflows using images and text
- Image generation and editing
- Audio recording and transcription
- Text-to-speech
- Embeddings and semantic search
- Workflows that save files, use extension storage, or show results without replacing text
- Extensions that need a custom dashboard instead of the generic workflow page

External workflow extensions also currently receive the same host-defined page. They can change titles, icons, commands, prompts, and feedback text, but cannot compose metrics, charts, settings, model pickers, shortcut editors, tables, or other dashboard elements.

## Goals

- Keep extension-specific workflows and dashboard definitions outside `nativ-core`.
- Let extensions compose functionality from stable, permission-checked host operations.
- Route model requests by task/capability rather than assuming a language model.
- Support typed values flowing between workflow steps.
- Allow safe custom dashboards without arbitrary in-process SwiftUI execution.
- Preserve ExtensionFoundation as the escape hatch for advanced native extensions.
- Validate packages completely before installation or activation.

## Proposed package structure

```text
MyExtension.nativextension/
├── Manifest.json
├── Workflow.json
├── Dashboard.json
└── Assets/
```

`Workflow.json` and `Dashboard.json` should remain optional and be referenced explicitly by manifest contributions.

## Typed workflow model

A workflow declares triggers and an ordered step graph:

```json
{
  "schemaVersion": 2,
  "triggers": [
    {
      "id": "describe-screen",
      "type": "shortcut",
      "shortcut": "Control+Shift+D"
    }
  ],
  "steps": [
    {
      "id": "capture",
      "type": "screen.capture"
    },
    {
      "id": "description",
      "type": "model.invoke",
      "task": "vision",
      "inputs": {
        "image": "{{capture.image}}",
        "prompt": "Describe this interface"
      }
    },
    {
      "type": "text.insert",
      "text": "{{description.text}}"
    }
  ]
}
```

[...]

PR #234, opened 2026-08-08
Restore extension permission controls and hide the Install button

Lazarus-931:
Solving the extensions page regressions noted in #230.

- Restore per-extension permission status and request controls in the Extensions page (status dots for Allowed / Denied / Not requested, tappable to request when the extension is enabled). These were dropped in the Hub redesign; the backing manager APIs were still intact.
- Hide the Install button until package installation ships (#163), as it was before.
- Remove the now-orphaned `ExtensionsView` (superseded by the Hub), moving its `FlowLayout` into the Hub.


Lazarus-931:
Screenshot 2026-08-08 at 8 46 41 PM

Blaizzy:
Thanks Alazar for the quick update!

Before we merge:

Let's follow this design for the card only, except the remove button and Nativ after the version. 

Because it's much easier to quickly distinguish between extensions and their content. As well as see which ones are included by default.

image

Blaizzy (approved):
LGTM, thanks!

PR #235, opened 2026-08-08
Bonsai Image model support

dcartman:
Prism ML has released a set of LLM and Vision models use 1 bit and Ternary inference. The version of mlx-vlm being used already has support for Qwen3 based LLM's, but lacks support for Image models. 

I realized this doesn't really belong in Nativ, it belongs in mlx-vlm, but I figured I would share my patch here in case anyone else wants to use Bonsai Image models.

HuggingFace: https://huggingface.co/prism-ml
Github: https://github.com/PrismML-Eng


Lazarus-931:
hi @dcartman, we'd love to support this, but I do agree this issue/feat request belongs in mlx-vlm!

feel free to close this and open there, will address it!

lucasnewman:
Just fyi the model is supported in `mlx-vlm` already (see https://github.com/Blaizzy/mlx-vlm/pull/1226) -- I think we just need the model type resolution fix there. Feel free to open a PR on `mlx-vlm` and we can make it work.

PR #240, opened 2026-08-09
Add live download statistics page

Blaizzy:
## Summary

- add a dedicated downloads page that presents live GitHub release-asset totals
- separate macOS installer downloads from automatic appcast update-feed requests
- support asset-type and release filters with shareable query parameters
- include a cached fallback snapshot when the GitHub API is unavailable
- add responsive desktop and mobile layouts plus homepage navigation links

## Why

Nativ's public download activity is currently split across installer and update-feed assets. This page makes the full history visible while clearly explaining that appcast requests are update checks rather than installations.

## Validation

- `node --check website/downloads.js`
- `node --check website/script.js`
- `git diff --check origin/main...HEAD`
- browser-tested live API loading and filter behavior
- visually verified desktop and mobile layouts


Lazarus-931:
nice, any pics?

Blaizzy:
Here you go!
Screenshot 2026-08-09 at 14 54 40

PR #242, opened 2026-08-09
Improve Models search and scrolling responsiveness

Blaizzy:
## Summary

Fixes the laggy, flickering Models experience—most noticeably when repeatedly scrolling up and down near the bottom of Discover—and removes the SwiftUI animation/layout loop that could keep Nativ near 85–95% CPU.

- keeps navigation and Installed/Discover switching responsive by isolating Models from unrelated app-wide runtime updates
- keeps search typing immediate while debouncing Installed filtering by 100 ms and network-backed Discover searches by 350 ms
- stabilizes Discover scrolling with lazy, identity-stable rows and row-local download progress updates
- removes layout-unstable header variants, scroll-edge control work, and automatic full metadata requests during scrolling
- stops hidden voice-capture overlays from running 60 fps animation timelines
- preserves the latest `main` navigation and Dev Hub changes while resolving the branch conflicts

## Root cause

The Discover page was repeatedly invalidating and recalculating SwiftUI layout while scrolling. App-wide model publications, multiple width-fitting header variants, scroll-edge button behavior, shared download progress, and hidden 60 fps voice-overlay timelines kept SwiftUI's animation/layout graph active even when the Models content had not changed.

Reversing scroll direction—especially near the bottom—amplified lazy-row recycling. That produced the lag and flicker that did not occur on the simpler Installed tab.

## Impact

Search text still updates immediately in the field. Installed filtering coalesces within 100 ms, while network-backed Discover searches wait 350 ms. Download progress redraws only its matching row, scrolling no longer triggers full Hugging Face metadata work, and hidden overlays no longer keep animation timelines alive.

The Models page remains lazy rather than being kept mounted, so the responsiveness improvement does not introduce a mounted-page memory tradeoff.

## Validation

- regenerated the Xcode project after merging the latest `main`
- completed a macOS Debug `xcodebuild` with code signing disabled
- stress-tested repeated up/down direction changes near the bottom of Discover without rows disappearing or flickering
- observed installed-app idle CPU settle around 9.5–11.8%, down from roughly 85–95%
- confirmed the fixed idle path no longer samples `ShapeStyleResolver`, `AnimatorState`, `TimelineView`, `ViewThatFits`, or popup-construction activity


lucasnewman (approved):
I ran it from your branch directly and it looks good! Nice!

PR #249, opened 2026-08-09
Simplify navbar navigation

Lazarus-931:
Simplifying navbar

Summary:
- Group Integrations and Developer under a collapsible Dev section.
- Keep the remaining navigation flat.
- Animate Dev rows through the leading edge in both directions.

https://github.com/user-attachments/assets/301a9786-7ee8-482d-8601-a8f8d62e17d8


Blaizzy (changes requested):
Collapsing these won't cut it.

I suggested we do a sub-page and sub-menu like the extensions page.

Lazarus-931:
seems like you are suggesting this

Screenshot 2026-08-10 at 9 30 16 AM

PR #250, opened 2026-08-10
Feature/browsing platform

Lazarus-931:
Add's browsing extensions, where users can input api keys for whatever search MCP provider they with and works out the box with models.


Blaizzy (changes requested):
I like the solution but there is one distinction to make.

Browsing is a tool not an extension. An extension is a custom interface and workflow that users/community/agent builds, and nativ's extension platform renders and executes.

Read more: https://github.com/Blaizzy/nativ/blob/main/Docs/Extensions.md

Please make the necessary adjustments.

Lazarus-931:
i see, will make it a part of tools

lucasnewman:
One minor thing: I think we should call the tool "web search" instead of "browsing", since it's more of an industry norm.

project.yml
```diff
@@ -175,6 +175,8 @@ targets:
       - path: Sources/Nativ/Features/Chat/Tools/ChatListModelsTool.swift
       - path: Sources/Nativ/Features/Chat/Tools/ChatServerStatsTool.swift
       - path: Sources/Nativ/Features/Chat/Tools/ChatSwitchModelTool.swift
+      - path: Sources/Nativ/Features/Chat/Tools/BrowsingSearchTool.swift
+      - path: Sources/Nativ/Features/Browsing/BrowsingProvider.swift
```
lucasnewman: Can you make sure to run `xcodegen` when adding files so the Xcode project builds out of the box?

Lazarus-931:
a to-do is to make this like a feature in the chat box, available from + dropdown

lucasnewman:
Looking good!

Two small tweaks I think that would help:
1) The title styling for web search is different than the other tools. I kind of like the more humanistic "Web Search" style, but I think just being consistent is good either way (up to you).
2) I think the tool should initially be off, and then when you flip it on it automatically brings up the settings so you can set an API key since all of the services require them. If you configure it, it stays on, and if you cancel it, it flips back off.

Screenshot 2026-08-10 at 12 28 05 PM

Issue #246, opened 2026-08-09
[Feature Request] Add audio file transcription and summarization


maya528:
What happened

Nativ currently does not provide a complete UI workflow for uploading an audio file and obtaining both a transcription and a summary.

Please add an audio file upload feature based on the mlx-audio framework.

Requested functionality

Upload local audio files, including .m4a and other common audio formats.

Support audio files up to 50 minutes in length.

Transcribe the uploaded audio locally.

Generate a summary based on the transcription.

Display the transcription and summary separately.

Show processing progress and clear error messages for unsupported or oversized files.

Keep the processing local and use the configured local models.

Expected behavior

After selecting an audio file, the user should be able to start processing and receive a complete transcription and summary without needing to use an external service.

<details>
<summary>Diagnostics</summary>

Environment

Nativ: 0.2.2 (202608041548)

macOS: macOS 26.6.1 (Build 25G76)

Chip: Apple M1 Pro

Memory: 32 GB total

mlx-vlm: 0.6.10

Speech-to-text model

Qwen/Qwen3-ASR-1.7B

</details>

Yavaren:
Fixed in #331

PR #253, opened 2026-08-10
Feature/audio file import

Lazarus-931:
for #246,

this allows users to upload a audio file and

1. run a stt model to transcribe it
2. play it


Lazarus-931:
closing since better designed #330 exists

PR #257, opened 2026-08-10
Fix debug signing and simplify hands-free shortcut

Blaizzy:
## Summary

- sign local Debug builds before opening them
- preserve microphone access while omitting profile-backed keychain sharing from profile-less Apple Development signatures
- restore hands-free modifier shortcuts to one tap to start and another tap to stop
- update the shortcut UI copy and state tests

## Root cause

The local signing script used a bare `return` after a failed Developer ID condition. Under `set -e`, that propagated a failure and stopped before the outer app was signed with its microphone entitlement. Once that was corrected, the profile-less Apple Development build still claimed a restricted keychain access group, causing macOS to terminate it at launch.

The double-tap requirement was introduced in response to issue #210, where the default modifier-only shortcut conflicted with macOS Spaces navigation. This change intentionally restores the original one-tap hands-free interaction: tap once to start and tap again to transcribe. Users who encounter a conflict can disable the global shortcut or assign a custom shortcut from the Audio shortcuts page.

## Impact

`make xcode-run` now builds, signs, verifies, and opens the exact Debug app. macOS can prompt that build for microphone access, and hands-free dictation uses one clean shortcut tap per toggle.

## Validation

- `make xcode-build`
- `make xcode-run`
- deep code-signature verification and successful launch of `dev.local.Nativ`
- focused shortcut tests were updated; their run is currently blocked by existing NativTests target errors for missing System Monitor sensor types

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .PHONY: build verify clean
-.PHONY: xcode-generate xcode-build xcode-run xcode-smoke xcode-lifecycle-smoke
+.PHONY: xcode-generate xcode-build xcode-sign xcode-run xcode-smoke xcode-lifecycle-smoke
 
 XCODE_DERIVED_DATA ?= build/NativDevelopmentDerivedData
 export DEVELOPER_DIR ?= /Applications/Xcode.app/Contents/Developer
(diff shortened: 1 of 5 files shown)


lucasnewman (approved):
LGTM

PR #262, opened 2026-08-10
Improve model discovery, downloads, and provider branding

Blaizzy:
## Summary

This PR improves the Models experience end to end: Discover results are easier to sort and more accurate, downloads report real activity and recover from stalls, memory warnings are clearer, and model publisher branding covers more of the catalog. It also fixes an adjacent voice-dictation shortcut regression.

## Discover

- Make **Trending · Descending** the default Discover ordering.
- Add explicit ascending and descending ordering, including stable local ordering across Nativ's five-page result window where the Hub API cannot provide ascending results.
- Organize capability filters into supported **Tasks** and **Features**.
- Send canonical Hugging Face pipeline filters for text, vision, audio, video, image generation/editing, speech-to-text, text-to-speech, and embeddings.
- Send reasoning, tool-calling, and drafter filters as Hub tags so they search the full catalog instead of filtering only a small page of unrelated trending results.
- Improve capability inference for supported Hugging Face task aliases and avoid presenting unsupported workflows as runnable models.
- Keep mixed SafeTensors/GGUF repositories discoverable while excluding GGUF artifacts from download size estimates and snapshot downloads.
- Add ascending/descending size sorting and simplify the Discover results header.

## Downloads and model cards

- Upgrade the bundled Hugging Face downloader stack to `huggingface_hub 1.27.0`, `hf-xet 1.6.0`, and `click 8.4.2`.
- Track monotonic progress from allocated cache bytes when downloader output is sparse.
- Show preparing, downloading, retrying, paused, and finalizing states; transfer speed is shown in the global download banner.
- Reserve 100% for completed downloads so finalization no longer appears stuck at 100%.
- Detect genuinely stalled transfers, retry automatically up to three attempts, and use a longer timeout while finalizing.
- Cleanly stop active downloader processes when Nativ exits.
- Replace the long system tooltip with a compact memory-fit popover that appears after an 80 ms hover and gives a clear smaller/quantized-model recommendation.

## Provider branding

- Add provider logos and resolution rules for InclusionAI, MiniMax, Baidu, Moonshot AI, Stability AI, Thinking Machines, and Meituan LongCat.
- Improve family matching for GLM, CogVLM/CogVideo, Kimi/Moonlight, ERNIE/Qianfan, Inkling, LongCat, and related variants.
- Resolve Meta-hosted and republished Muse-Glimmer variants to the Meta logo, including MLX Community, Unsloth, and custom fine-tunes.

## Voice shortcut fix

- Require a complete modifier release before the Fn + Control toggle can re-arm, preventing a stop action from immediately starting a new dictation when NSEvent and the polling fallback deliver stale overlapping samples.

## Maintenance

- Keep the Xcode project source membership synchronized for `ChatWorkspacePicker.swift`.
- Add regression coverage for capability mapping, GGUF exclusion, provider resolution, download progress/finalization/speed, stall activity, and the modifier-toggle state machine.

## Root cause

Discover previously applied some capabilities only after fetching a small trending page, so reasoning and tool-calling results were incomplete or inaccurate. The download UI relied too heavily on sparse subprocess progress, which made Xet transfers appear frozen at 0% or 100% and offered no bounded recovery when a process actually stalled. Provider resolution also leaned on repository owners, so republished model families could lose the correct logo. Finally, duplicate modifier samples could re-arm the voice toggle before every shortcut key was released.

## Validation

- 39 focused model/provider/download/shortcut tests passed with 0 failures.
- `make xcode-run`
- Rebuilt and signed the Debug app with an Apple Development identity.
- Verified the app signature and launched the resulting bundle successfully.
- `git diff --check`

## Preview

Compact model memory warning

Screenshot 2026-08-12 at 02 53 46


Lazarus-931 (approved):
all good to me, just I think instead of scattering the model providers everywhere, we can keep them in a single struct which holds their svg, monogram, name, etc like:

```Swift
struct ProviderBrand {
    let displayName: String
    let iconResourceName: String?
    let monogram: String
    let preservesIconColors: Bool
    let needsLightBackgroundInDarkMode: Bool
    let tint: NSColor
    let organizationKeys: [String]
    let familyIdentifiers: [String]
}
```

doesn't have to be swift although that'd be best, can be json of model providers

Also we should include x.ai and bytedance!

Blaizzy:
>  just I think instead of scattering the model providers everywhere

What do you mean? Could you be more specific as to where the model providers are scattered? 

We only use one file `.../Models/LocalModelProvider.swift` and in ModelsView we just extend it `extension LocalModelProvider {`. If you are talking about the extension, I can refactor that into `LocalModelProvider.swift`. And if you are referring to refactoring `LocalModelProvider`, I can do that too.

Lazarus-931:
> > just I think instead of scattering the model providers everywhere
> 
> What do you mean? Could you be more specific as to where the model providers are scattered?
> 
> We only use one file `.../Models/LocalModelProvider.swift` and in ModelsView we just extend it `extension LocalModelProvider {`. If you are talking about the extension, I can refactor that into `LocalModelProvider.swift`. And if you are referring to refactoring `LocalModelProvider`, I can do that too.

yes, those two would be nice to refractor

PR #264, opened 2026-08-11
Add Parallel Search to the MCP catalog

georgeatparallel:
Nativ's MCP catalog gives local models access to additional tools without changing the app's defaults. This adds Parallel Search as an optional entry for live web search and clean URL fetching through its free remote MCP endpoint.

## What changed

- Added a `parallel-search` catalog entry using the existing `npx` launcher pattern.
- Pinned `mcp-remote` to `0.1.38` and selected its Streamable HTTP transport.
- Kept the entry credential-free and left the existing Fetch server and native web-search providers unchanged.

## Validation

- Parsed `MCPCatalog.json` successfully.
- Passed `git diff --check` and confirmed the patch only appends the approved catalog record.
- Passed `python scripts/verify_mcp_catalog.py --only parallel-search` with two tools.
- Confirmed the endpoint lists `web_search` and `web_fetch` in a credential-free temporary environment.

Disclosure: I work at Parallel.ai, which operates this MCP endpoint.

### DIFF
--- a/Sources/Nativ/Resources/MCPCatalog.json
+++ b/Sources/Nativ/Resources/MCPCatalog.json
@@ -59,5 +59,14 @@
     "symbol": "cylinder.split.1x2",
     "tint": "green",
     "sourceURL": "https://github.com/modelcontextprotocol/servers/tree/main/src/sqlite"
+  },
+  {
+    "id": "parallel-search",
+    "name": "Parallel Search",
+    "summary": "Search the live web and fetch clean content from URLs.",
+    "command": "npx",
+    "args": ["-y", "mcp-remote@0.1.38", "https://search.parallel.ai/mcp", "--transport", "http-only"],
+    "symbol": "globe",
+    "tint": "indigo"
   }
 ]


Lazarus-931:
thanks @georgeatparallel, tha'd be a great addition to nativ!

just some nits:

does parallel have curl endpoints we can wire into the existing Web Search tool? if so, I'd think it makes sense parallel becomes one of many options people can use with api key, see this section below:

https://github.com/Blaizzy/nativ/blob/b314e683daea2ef56d56f5ee5598b538a786bc20/Sources/Nativ/Features/Browsing/WebSearchProvider.swift#L3

if so, then id go with that since it should conform to how other search api providers are set. there is a extensions tool with different providers like brave, exa, perplexity, etc which I think would be a place best for this pr, along with it's logo, and heuristics. with it, also the right endpoints to hit for search 

but let me know what you think, I'm not too familiar with how parallel best works :)

Blaizzy:
Thanks for the contribution @georgeatparallel!

georgeatparallel:
> thanks @georgeatparallel, tha'd be a great addition to nativ!
> 
> just some nits:
> 
> does parallel have curl endpoints we can wire into the existing Web Search tool? if so, I'd think it makes sense parallel becomes one of many options people can use with api key, see this section below:
> 
> https://github.com/Blaizzy/nativ/blob/b314e683daea2ef56d56f5ee5598b538a786bc20/Sources/Nativ/Features/Browsing/WebSearchProvider.swift#L3
> 
> if so, then id go with that since it should conform to how other search api providers are set. there is a extensions tool with different providers like brave, exa, perplexity, etc which I think would be a place best for this pr, along with it's logo, and heuristics. with it, also the right endpoints to hit for search
> 
> but let me know what you think, I'm not too familiar with how parallel best works :)

That’s my understanding too: this PR adds the free Search MCP as a lightweight catalog option, while the native provider route would be a deeper API-key integration in Nativ’s built-in Web Search tool. I’m happy to pursue that. Would you prefer I reshape this PR around the native provider, or keep the zero-setup MCP option and send the native provider as a focused follow-up?

Lazarus-931:
> > thanks @georgeatparallel, tha'd be a great addition to nativ!
> > just some nits:
> > does parallel have curl endpoints we can wire into the existing Web Search tool? if so, I'd think it makes sense parallel becomes one of many options people can use with api key, see this section below:
> > https://github.com/Blaizzy/nativ/blob/b314e683daea2ef56d56f5ee5598b538a786bc20/Sources/Nativ/Features/Browsing/WebSearchProvider.swift#L3
> > 
> > if so, then id go with that since it should conform to how other search api providers are set. there is a extensions tool with different providers like brave, exa, perplexity, etc which I think would be a place best for this pr, along with it's logo, and heuristics. with it, also the right endpoints to hit for search
> > but let me know what you think, I'm not too familiar with how parallel best works :)
> 
> That’s my understanding too: this PR adds the free Search MCP as a lightweight catalog option, while the native provider route would be a deeper API-key integration in Nativ’s built-in Web Search tool. I’m happy to pursue that. Would you prefer I reshape this PR around the native provider, or keep the zero-setup MCP option and send the native provider as a focused follow-up?

yeah totally think we can merge this, feel free to open a pr if you ever want parallel to be supported along with other search providers in Web Search tool!

Issue #260, opened 2026-08-10
Models not available configuring OpenCode Integration


nativ260810:
Today I downloaded and installed Nativ Version 0.2.2 (202608041548).
But I find I can't make use of the OpenCode integration with one of the models I previously downloaded.

## Preparation
From the **Models** page I selected `+ Add Folder ...` and added my `~/.lmstudio/models` and `~/.omlx/models` directories.
I observe several models as expected.  They all have an "External" capsule.
I selected one and observed the momentary "Loading" capsule.
From the **Chat** page I asked a trivial question and observed a reasonable response.

## Problem
From the **Integrations** page among all the options I observe "OpenCode ... Ready to configure"
When I select OpenCode I see

<img width="773" height="142" alt="Image" src="https://github.com/user-attachments/assets/03d7fb52-f51d-4a24-aa10-d659466c86fd" />

None of the models shown on the models page are available and the `Configure` and `Configure & Open` buttons are gray.

## Question
How can I configure the OpenCode integration to use a model I previously downloaded which works fine from **Chat**?

I know the model in question works just fine from OpenCode when served by LM Studio and oMLX.
I don't want to download it again just for this.

Lazarus-931:
hey @nativ260810 , have a fix in #226, once merged, models outside of hf cache folder should be picked up!



<img width="895" height="864" alt="Image" src="https://github.com/user-attachments/assets/c4dcaec5-10dd-4471-b0c4-56768926d959" />

nativ260810:
> hey @nativ260810 , have a fix in https://github.com/Blaizzy/nativ/issues/226, once merged, models outside of hf cache folder should be picked up!

Did you mean to write **https://github.com/Blaizzy/nativ/issues/266** ?  That looks more relevant.
Looks like fix isn't in the latest release https://github.com/Blaizzy/nativ/releases/tag/v0.3.0 
Once a release with it is ready I'll try again.
Thank you @Lazarus-931 for the prompt followup!

PR #266, opened 2026-08-11
Set up global model folder access

Lazarus-931:
Centralizes model discovery so folders added in Models are available across Nativ’s tools, actions, utilities, and integrations, addressing [#260](https://github.com/Blaizzy/nativ/issues/260).

I'd say prior to this, this was poorly designed. Additional folders alongside main one should be available across all components of Nativ that might need model folders, done so by defining a `LocalModelSearchPaths(primary: modelSearchPath, additional: additionalModelSearchPaths)`


lucasnewman (approved):
LGTM

PR #269, opened 2026-08-11
Fix unknown-tool error mislabeling in the tool dispatcher

C-K-Loan:
An unrecognized tool call (hallucinated name, or a typo) was mislabeled as an image-tool failure instead of a clear "unknown tool" error, since `ChatToolDispatcher` had no dedicated case for it and fell through to `ChatImageToolError.unsupportedTool`.

Adds `ChatUnknownToolError`/`ChatUnknownToolResultPayload` (neutral `ok`/`error` shape, no `operation` field) so an unrecognized name now surfaces as `"Unknown tool: <name>"` instead of impersonating an image-tool error.

2 files, +51/-2. 525/526 tests pass; the one failure (`testBuiltInToolsUseThePresentationOrder`) is a pre-existing, unrelated tool-order drift on current `main`, not touched here.

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatToolRegistry.swift
+++ b/Sources/Nativ/Features/Chat/ChatToolRegistry.swift
@@ -193,6 +193,22 @@ enum ChatToolRegistry {
     }
 }
 
+enum ChatUnknownToolError: LocalizedError {
+    case unknownTool(String)
+
+    var errorDescription: String? {
+        switch self {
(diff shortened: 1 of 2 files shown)


lucasnewman:
@C-K-Loan Thanks! Can you fix the last conflict and we can merge?

C-K-Loan:
Sure @lucasnewman conflicts resolved ready to merge

PR #271, opened 2026-08-11
Gate integration discovery behind an explicit action

brainx:
## Summary

- Stop the Integrations page from scanning configured model folders or probing installed tools on page appearance and model-library notifications.
- Add an explicit **Check integrations** action, with **Refresh** as the repeat action.
- Disclose the local operations performed by discovery and add accurate progress, disabled, and accessibility states.
- Preserve the global model search paths introduced in #266 and cover passive lifecycle events with regression tests.

## Why

Opening **Dev → Integrations** currently starts local discovery without a direct user action. Discovery can scan configured model folders, run the login shell, read known integration configuration files, and execute installed tools with `--version`.

This change makes those operations visible and user initiated. It is intentionally scoped to discovery initiated by the Integrations page: it does not start an interactive agent session, choose or change a project folder, alter app-wide model scanning elsewhere, or sandbox external tools.

## Checks

- `git diff --check upstream/main...HEAD`
- Focused `IntegrationServicesTests` and `LocalModelDiscoveryTests`: 40 passed, 0 failed
- Nativ macOS app build passed using a temporary XcodeGen project generated from `project.yml`; only the unrelated downloadable `mlx-vlm-server` resource packaging phase was omitted

### DIFF
--- a/Docs/features/integrations.md
+++ b/Docs/features/integrations.md
@@ -4,6 +4,11 @@ Nativ serves local models over standard APIs, so external agents and editors tha
 Anthropic can point straight at it. The **Integrations** page writes each tool's configuration
 automatically; the per-tool sections below document what each one needs for manual setup.
 
+The **Integrations** page does not start model or tool discovery when opened. Choose **Check
+integrations** to scan the configured model folders, run the login shell, read known tool
+configuration files, and execute installed tools with `--version`. The check does not start an
+interactive agent session or choose a project folder; **Refresh** repeats it explicitly.
+
 ## Connection details
 
 - OpenAI-compatible base URL: `http://127.0.0.1:8080/v1`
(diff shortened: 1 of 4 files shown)


Lazarus-931:
hi @brainx, thanks for the pr!

sry, but i'm quite unsure how this is useful. 

#266 is under the assumption people select directories they want Nativ to access, and making those additional directories shared access across different Nativ calls 

Could you explain it for me please?

Blaizzy:
Hey @brainx! Happy to clarify.

We check for a predefined set of tools so Nativ knows what's installed and configured on your machine, and can surface the right integrations to you. It only looks for that specific list, nothing else is scanned, collected, or sent anywhere.

It works exactly like our model detection: we check what models you have locally so we can show them in the app. Same principle here.

Since nothing leaves your machine and no personal data is touched, we don't plan to add a separate consent prompt for it. But if you have concerns about how it works, I'm all ears, always open to feedback on this stuff.

brainx:
Hi, this was an older PR idea for that I acted on too late. By then, #266 had already covered the directory side, and I mixed the original idea with the separate integration-check behavior. That made this PR redundant. Closing it makes sense—appreciate the clarification.  👍

PR #274, opened 2026-08-11
Stabilize onboarding model metadata loading

Lazarus-931:
see in #265, concurrent fetch requests, where for each model a worker is assigned, which led to crashes.

this pr introduces `HuggingFaceCuratedModelLoader` which fetches first, and return raw Data from Hugging Face instead of decoding inside each network task. This is followed by a bounded network concurrency, which only allows max of 4 requests, launched one after another. 

After all arrive a single `JSONDecoder` processes them, which all in all is an effort to not run more than one `HuggingFaceModel` decode concurrently.

Also, failed requests are skipped, along with invalid JSON.

Verification with main could only be that this pr doesn't crash when started from onboarding, unlike main

### DIFF
--- a/Sources/Nativ/Features/Models/HuggingFaceHub.swift
+++ b/Sources/Nativ/Features/Models/HuggingFaceHub.swift
@@ -421,7 +421,7 @@ private struct HuggingFaceHubClient: Sendable {
         }
     }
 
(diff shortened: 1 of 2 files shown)


lucasnewman (approved):
🙏

PR #275, opened 2026-08-11
Generation only models present an error when a context image is added to user message.

dcartman:
Bonsai image models are generation-only models (text-to-image) and are unable to edit images (image-to-image). Currently if a context image is included with the users message a not-helpful error is presented to the user. This changes prohibits the user from submitting a context image to a generation-only model.

AI generated code, human reviewed.


Lazarus-931:
hi @dcartman, so I like this pr! 

I definitely think this is a needed component, and instead encourage you to do a pr upstream in mlx-vlm.

I don't think

```Swift

  if let selectedModel = imageModels.first(where: {
      $0.repoID == viewModel.modelID
  }) {
      return selectedModel.capabilities.contains(.imageEditing)
  }

  return MLXImageModelResolver.isKnownImageEditOnlyModelID(viewModel.modelID)

```

is necessarily the right approach, since now, models that are Unknown get blocked.

Instead, in mlx-vlm, we should expose the existing `is_image_edit_model(model)` function so Nativ can pick it up!

That would mean exposing it through it's endpoint like:

```JSON
{
    "data": [
      {
        "id": "/local/snapshot/path",
        "capabilities": ["image_generation", "image_editing"]
      }
    ]
 
```

then Nativ would cache and enable image_editing!

I'd recommend closing this and opening one with this similar approach, happy to answer any questions along the way though!

dcartman:
Yeah, I agree mlx-vlm should be the source of truth for model capabilities.

I'm working on a pull request for mlx-vlm that will expose the model capabilities through the /models end point. I'll link to that pull request once it's up.

Any objection to using this pull request to track the nativ changes required to adopt the capabilities object? e.g. blocking the user from submitting a context image to a generation-only model.

Lazarus-931:
awesome! 

not a problem, we can keep this open for you to make changes there and then edit this to be compatable

dcartman:
Here is the pull request for the `mlx-vlm` work: 
https://github.com/Blaizzy/mlx-vlm/pull/1859

I'll update this pull request shortly with the nativ client work.

PR #277, opened 2026-08-11
Polishing UI

Lazarus-931:
- Color change
- Removing un-needed text


Yavaren:
Noted, won't touch this

PR #280, opened 2026-08-11
Stabilize Kits UI layout and status icon

Yavaren:
- replace the Enabled and Needs setup badges with a green completion indicator
- keep Kit cards at stable, equal dimensions across state changes
- prevent card content, badges, and controls from shifting
- made spaces inside the cards stable across all states

The previous iteration had issues with the overall layout of cards, they changed in size and created empty spaces based on the state of toggles.

image
image


Lazarus-931:
thx @Yavaren

PR #281, opened 2026-08-12
Working on global MCP tool skill kit and extent and extension management

Lazarus-931:
Working on global MCP, tool, skill, kit, and extension management, as well as Web Search and Deep Research capabilities.


Sources/Nativ/Features/Shared/NativComponents.swift
```diff
@@ -1,5 +1,178 @@
+import AppKit
 import SwiftUI
 
+struct NativArrowlessPopoverPresenter<Content: View>: NSViewRepresentable {
```
lucasnewman, line 4: Could we use a normal popover here? What's the limitation in SwiftUI that prevents us from using the [standard implementation](https://developer.apple.com/documentation/swiftui/view/popover(ispresented:attachmentanchor:arrowedge:content:))?

Blaizzy: We can't because of design intent
image

PR #282, opened 2026-08-12
Animate model loading through the Nativ mark

Lazarus-931:
Nativ logo appears when model is loading, instead of the blue loading UI that was out of place.

Logo appears when model is loaded

https://github.com/user-attachments/assets/4cd923d7-0ed4-4e12-901e-ca2246a58574


Blaizzy:
This looks cool! 

Tbh though, more animations aren't the direction we want to take, we lean toward keeping the UI fast and out of your way. We already display the model loading status in 3 places in that window, so another one would be redundant. What would help a ton is UX fixes and improvements, that's where the real impact is right now. Happy to point you at a few areas if you're up for it 👌🏽

PR #285, opened 2026-08-12
improving markdown rendering in chat

Lazarus-931:
Assistant replies streamed as raw text (visible `##`, `-`, and ```) and only snapped into styled markdown once the response finished. Now the streaming buffer goes through the same renderer, with an unterminated code fence closed virtually so code blocks render styled as they arrive.


Lazarus-931:
https://github.com/user-attachments/assets/4f4c31a5-5c25-40da-8d14-f2be2099f046

Blaizzy (approved):
LGTM, thanks and well done!

Blaizzy:
Screenshot 2026-08-13 at 03 12 07

Spikes CPU and slows down generation to 1/3, and the PC freezes

Lazarus-931:
will patch with a follow up pr

edit: seems like sending the entire streaming buffer through the full StructuredText pipeline was bad choice

Lazarus-931:
though it seems this needs a careful approach, i think for now we can revert this feature and add it with better design

PR #286, opened 2026-08-12
Recover from a frozen chat instead of locking the UI

Lazarus-931:
polishing how frozen chat is handled, had a case where if it freezes:

1. can't delete the chat b/c it's now still "loading"
2. can't start another chat
3. can't pause the model

to recreate, start server and ask model for long response, as it writes, end the server, any text written from then on freezes the entire app, only option is to quit it.

The in-flight request slot is now released on every exit path instead of only the happy one, so a stream that stalls can no longer leave a chat permanently "generating."

### DIFF
--- a/Sources/Nativ/NativModel.swift
+++ b/Sources/Nativ/NativModel.swift
@@ -46,6 +46,7 @@ final class NativModel: ObservableObject, ChatModelSwitchingSurface {
 
     @Published private(set) var modelSwitchInProgress = false
     @Published private(set) var modelSwitchTargetID: String?
+    private var modelSwitchWatchdog: Task<Void, Never>?
     @Published private(set) var modelLoadingProgress: Double?
     @Published private(set) var modelLoadFailure: ModelLoadFailure?
     @Published private(set) var modelPreloadMemoryWarning: ModelPreloadMemoryWarning?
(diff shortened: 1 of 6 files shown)


Sources/Nativ/NativModel.swift
```diff
@@ -1012,4 +1016,35 @@ final class NativModel: ObservableObject, ChatModelSwitchingSurface {
...
+    /// started server. When that server comes up but never serves metrics, the
+    /// flag used to stay set forever, disabling the model picker and the
+    /// start/stop buttons with no way to recover short of relaunching. The
+    /// watchdog re-checks the flag and target before acting, so a stale one that
+    /// fires after a successful switch is a no-op.
+    private func armModelSwitchWatchdog(
+        timeout: TimeInterval = NativModel.modelSwitchTimeout
+    ) {
+        let targetID = modelSwitchTargetID
+        Task { @MainActor [weak self] in
```
lucasnewman: Do we need to hold a reference to this task and cancel it in case multiple model switches happen in a row?

Lazarus-931: fixed!

lucasnewman (approved):
LGTM

PR #294, opened 2026-08-13
Revert the streaming markdown rendering

Lazarus-931:
Reverts #285, which is in 0.3.1. Rendering the whole message through the markdown renderer on every stream flush (15/sec) pins a core and slows generation, worst with tables. Reverting to the inline path while the performance is fixed properly.


Blaizzy:
No needed, I fixed performance whilst keeping the feature  ✅

PR #296, opened 2026-08-13
rerouting user if image model is now downloaded

Lazarus-931:
finishes the tasks from #131, now routes user when chat model is trying to tool-call a image model, when none is present:

Screenshot 2026-08-13 at 9 26 55 AM


Yavaren:
@Lazarus-931 I had a PR regarding this problem, what is new about this solution? I thought I solved it already

Yavaren:
This is the current UI on main 

It suggests models, but you can get browse models as well 

image

Yavaren:
But I agree that we should better explain that no compatible model is downloaded

Lazarus-931:
my mistake, assumed not so since issue was still up, closing #131

Yavaren:
@Lazarus-931 you had a good idea of showing that no compatible model installed, I did not think about it. I will update the feature. Thanks!

PR #298, opened 2026-08-13
Fix model download progress reporting and added progress bar

### LINKED ISSUE #232: Model downloads can be extremely slow or remain stuck indefinitely
## Summary

Model downloads can be unexpectedly slow and, in some cases, appear to remain stuck indefinitely. The download may stay in an active/downloading state without meaningful progress, leaving the model unavailable and requiring the user to cancel or restart the app.

## Steps to reproduce

1. Open the Models view.
2. Start downloading a model.
3. Observe the download progress over time.
4. Repeat with another model or after restarting the app if the first download becomes stuck.

## Actual behavior

- Download speed can be much slower than expected.
- Some downloads stop making visible progress and do not complete.
- There is no clear indication whether the download is still active or stalled.
- The user is not given an actionable timeout, error, retry, or resume path.

## Expected behavior

- Show accurate progress, transfer speed, and/or an ETA.
- Detect when a download has made no progress for a reasonable period.
- Surface a clear, actionable error when a download stalls.
- Allow the user to retry, resume, or cancel cleanly without leaving the model in a misleading partial state.
- Preserve useful diagnostic information for troubleshooting network or server failures.

Yavaren:
Closes #232

- report responsive byte-level model download progress
- derive transfer speed from actual Xet transfer events
- preserve the existing download progress UI

Screenshot 2026-08-13 at 5 06 20 PM


lucasnewman (approved):
👌

PR #306, opened 2026-08-14
Reduce streaming chat CPU usage

Blaizzy:
## Summary

- Isolate the active assistant stream state from the published transcript.
- Append streaming text incrementally in an AppKit text view.
- Throttle stream-follow scrolling while preserving token-level updates and final Markdown rendering.

## Root cause

Each streamed delta mutated the full published message array, invalidating the complete SwiftUI chat hierarchy, including composer controls and accessibility state. The cost grew with response length and eventually saturated a CPU core, which also reduced model decode throughput.

## Impact

Streaming keeps the existing UX, but updates remain local to the active response. Completed content is committed once and rendered as Markdown as before.

## Validation

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make xcode-build`
- `NATIV_TEAM_ID=WT753TG39U DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make xcode-run`
- Long-response benchmark: CPU stayed roughly 35–50% instead of climbing toward 100%; decode throughput remained around 30 tok/s.
- Recursive signature verification passed.

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatView.swift
+++ b/Sources/Nativ/Features/Chat/ChatView.swift
@@ -145,7 +145,8 @@ private struct ChatTranscriptView: View {
                     ForEach(chat.visibleMessages) { message in
                         let editUnavailableReason = userPromptEditingUnavailableReason(for: message)
                         ChatMessageRow(
-                            message: message,
+                            sourceMessage: message,
+                            streamState: chat.streamState(for: message),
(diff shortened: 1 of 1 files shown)


lucasnewman:
This is obsoleted by https://github.com/Blaizzy/nativ/pull/308

PR #307, opened 2026-08-14
Restrict the state observed by the sidebar to reduce steady state relayout

lucasnewman:
The monolithic NativModel observation is kind of an anti-pattern in SwiftUI, so I'm incrementally untangling the other views from it. This updates the sidebar to only observe properties it needs to render, which brings its steady state rendering down to zero SwiftUI View updates instead of a constant trickle.


Blaizzy (approved):
LGTM, thanks!

PR #309, opened 2026-08-15
Feature/shared document extraction

Yavaren:
Adds PDF support to chat using the shared document extraction service.

- Text-only models can process PDFs without requiring vision support.
- Images remain restricted to vision-capable models.
- Artifacts and Chat now share the PDF extraction implementation.
- PDF text is extracted once, cached in memory, and added to model requests without being persisted.
- Context is limited to 24,000 characters per PDF and 48,000 per request.
- Added clear warnings for unsupported, unreadable, scanned, password-protected, and truncated documents.
- Renamed **Upload Image…** to **Upload Files**.

image
image


Sources/Nativ/Features/Chat/ChatAttachmentNotice.swift
```diff
@@ -0,0 +1,32 @@
+import Foundation
+
+struct ChatAttachmentNotice: Identifiable, Equatable {
```
lucasnewman: Can we just inline this file with the view class?

Sources/Nativ/Features/Chat/ChatAttachmentValidation.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+enum ChatAttachmentValidation: Equatable, Sendable {
```
lucasnewman: Let's inline this file with the validator.

Sources/Nativ/Features/Documents/DocumentTextExtracting.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+protocol DocumentTextExtracting: Sendable {
```
lucasnewman: If we're only supporting PDF I don't think we need the protocol here. I'd rather refactor it later when targeting a new document type vs adding a bunch of complexity now for an abstraction that we don't actually need. I think we can also just inline all three of these files into one file -- having tiny files that define a single type makes it harder to navigate the project, not easier.

Sources/Nativ/Features/Chat/ChatAttachmentNotice.swift
Yavaren: Yes, agreed

Sources/Nativ/Features/Chat/ChatAttachmentValidation.swift
Yavaren: Yes, agreed

Sources/Nativ/Features/Documents/DocumentTextExtracting.swift
Yavaren: We had a long discussion regarding other file formats, but we'll leave it for later, and it seems like the best direction for now is to only extract text from PDFs. I started preparing the repo for multi-format extraction, but I agree that let's leave it lean for now until we come back to this question.

lucasnewman: Yep, I think in general your abstraction will be correct when we support more file types but let's cross that bridge when we come to it :)

Yavaren: Thanks a lot for your feedback. I pushed a commit, ready for your review :)

Yavaren: Not yet, I have very slow internet right now, did't see new merge conflicts. Will be ready in 5 min.

Yavaren: @lucasnewman done

PR #312, opened 2026-08-15
Protect routine and chat lifecycle

Lazarus-931:
Before, chat sessions and routines were managed separately, so deleting a chat could leave its routine and scheduled launch agent behind. This adds shared lifecycle cleanup so deleting a chat also removes its associated routine, including through bulk deletion.

Also, adds cancellation and validation to queued or active routine runs, so when deleted, routines cannot keep running or recreate chats and run history after completion .

Startup reconciliation now removes routines whose source chats no longer exist.

Also, routines cannot silently outlive their chats, and the new tests cover deletion, cancellation, orphan cleanup, and stale completion cases.

This is draft and a temp fix, but long term, i think we should have a `SessionStore`  aggregate design with an optional routine, and the entire lifecycle centralized

Closes #300


Lazarus-931:
cc: @Yavaren

Blaizzy:
This inverts the relationship. Routines produce chats; chats don't own routines. Deleting a chat should be like deleting a GH Actions run, the log goes, the workflow stays.

PR #313, opened 2026-08-15
Move ChatViewModel to its own file and add tests

lucasnewman:
ChatViewModel has grown to 2k+ LoC with tools and other features, and had effectively no direct test coverage. This breaks it out into its own file and adds a basic test suite.

With this change ChatView.swift is back down to 1.7k LoC from nearly 4k which makes it significantly more readable. No functionality change, this is just mechanical.


Lazarus-931 (approved):
much needed, thx @lucasnewman

PR #314, opened 2026-08-15
Clean up built-in tools listing

lucasnewman:
1) Use a user-friendly tool description instead of the actual tool instructions.
2) Remove the duplicated "Built-in" badge from built-in tools.
3) Allow disabling built-in tools, as some small models (e.g. Qwen3-0.8B) will often make poor/irrelevant tool calls when present.
4) Reorganize the listing to group similar tools.
5) Only show the Web Search and Web Read tools as enabled if they are actually configured, and show the configuration sheet when enabling them.

Screenshot 2026-08-15 at 3 12 06 PM


Lazarus-931 (approved):
👍🏽

PR #315, opened 2026-08-17
Feature/live server settings

Lazarus-931:
This pr introduces a v1/settings/ endpoint support for engine, with it's own section for GET and PATCH endpoint. 

 https://github.com/Blaizzy/mlx-vlm/pull/1876 has been merged, so this should be gtg for when we do a mlx-vlm release.

Screenshot 2026-08-17 at 9 09 11 AM


Lazarus-931:
@Yavaren, works fine for me with latest mlx-vlm pinned to local build, can u also re-verify design + integration gtg?

lucasnewman:
@Lazarus-931 Can you put up an updated screenshot since the mlx-vlm changes are in?

Lazarus-931:
> @Lazarus-931 Can you put up an updated screenshot since the mlx-vlm changes are in?

Screenshot 2026-08-19 at 3 29 06 PM

lucasnewman:
The massive list feels pretty intimidating. How about breaking these down into disclosure groups with related functionality, and hiding irrelevant switches if the top-level feature switch is off?

I also think freeform entry vs a dropdown of supported values for some of these could do more harm than good, e.g. what would specifying 1 for "KV Bits" do? It's not supported, so I presume it will just error out every request or crash the server.

Lazarus-931:
> The massive list feels pretty intimidating. How about breaking these down into disclosure groups with related functionality, and hiding irrelevant switches if the top-level feature switch is off?
> 
> 
> 
> I also think freeform entry vs a dropdown of supported values for some of these could do more harm than good, e.g. what would specifying 1 for "KV Bits" do? It's not supported, so I presume it will just error out every request or crash the server.

yes, your quite right. i'm working on improving it, that was just what i last had built

Lazarus-931:
Screenshot 2026-08-22 at 6 12 38 PM

Blaizzy:
Reverting this because it needs a proper server settings redesign and a few other reasons:

**Styling**
- Buttons and the text box don't fill the available width; they're noticeably smaller than the rest of the page.
- In full-screen, the container doesn't match the width of the other sections.
- The overall layout is hard to navigate.
- Copy saying “changes take effect without restarting the server”, doesn’t help the user. 

**Dead configs**
- Speculative decoding and prefix caching are exposed but not wired up.

**Duplication**
- Most of these settings already live in the right-hand panel, so this largely repeats what's there.

Happy to sketch out what a redesign could look like in a spec file if that helps.

## Minimized
Screenshot 2026-08-25 at 10 31 28

## Fullscreen
Screenshot 2026-08-25 at 10 37 35

Blaizzy:
## Existing model configuration panel 
Screenshot 2026-08-25 at 10 40 11

PR #316, opened 2026-08-17
improve xet download progress bar UI

Yavaren:
- Keep Xet download progress moving between reconstruction updates.
- Reserve the final 5% for work Xet cannot measure precisely.
- Show a clear “Finishing…” state with an indeterminate progress bar.
- Explain that downloaded model files are being assembled.

## Problem

Xet reports network transfer and model-file reconstruction separately. Reconstruction updates can arrive in bursts, making progress appear stuck even while work continues.


lucasnewman:
@Yavaren I made a minimal fix for now in https://github.com/Blaizzy/nativ/pull/317 as I believe there are still some issues here:

`let estimatedLimit = Int64(Double(totalBytes) * 0.95)`

This truncates downward, while isFinishing requires the fraction to be >= 0.95, so the cap can get stuck at 0.9499… and never enter “Finishing.” until another reconstruction update arrives. You should rebase your changes onto the minimal fix and we can incorporate the UI changes.

lucasnewman (approved):
LGTM

PR #322, opened 2026-08-17
Fix audio transcription configuration lifetime

### LINKED ISSUE #321: Audio recording reports that the Nativ server is not running and shows incorrect title. Both during audio transcription recording.
## Description

After completing an audio recording, the app displays the following error:

> Start the Nativ server before transcribing or summarizing a recording.

The server is running and healthy, and a speech-to-text model is installed and selected. No transcription request reaches the server.

## Steps to reproduce

1. Enable the Audio extension.
2. Start the Nativ server.
3. Confirm that a speech-to-text model is installed.
4. Open the Audio page.
5. Start and complete a recording.

## Actual result

The recording remains in the saving and transcribing state before the app displays an incorrect “server not running” error.

## Expected result

The saved recording should be sent to the running server for transcription.

## Technical details

Debugger inspection confirmed that the Audio extension still has a transcription configuration provider, but the provider returns `nil`.

The provider weakly captures `AppDelegate`. When that captured delegate instance is no longer available, the closure returns `nil`, which `AudioCaptureLibrary` incorrectly interprets as the server being stopped.

The provider should retain the `NativModel` dependency it reads instead of depending on the lifetime of `AppDelegate`.

## Additional UI issue

While recording, the active capture panel displays **“Recording recording.”**

This happens because the UI prefixes the active record kind with “Recording,” while the record kind itself is also titled “Recording.” The status should display simply **“Recording.”**

Image
Image

Yavaren:
Closes #321 

## Summary

- Keep the live `NativModel` available to the Audio extension’s transcription configuration provider.
- Prevent completed recordings from incorrectly reporting that the Nativ server is stopped.
- Simplify the active recording status from “Recording recording” to “Recording.”

## Root cause

The transcription configuration provider weakly captured `AppDelegate`. The Audio extension could retain the provider after the captured delegate instance was released.

In that state, the provider returned `nil`, which `AudioCaptureLibrary` interpreted as the server being stopped.

The duplicated status was produced by combining the “Recording” state prefix with an active record kind whose title was also “Recording.”

image
image

### DIFF
--- a/Sources/Nativ/Features/Extensions/ExtensionsHubView.swift
+++ b/Sources/Nativ/Features/Extensions/ExtensionsHubView.swift
@@ -40,13 +40,6 @@ struct ExtensionsHubView: View {
         .task {
             guard !didLaunch else { return }
             didLaunch = true
-            manager.launch(
-                context: NativExtensionHostContext(
-                    transcriptionConfiguration: { nil },
-                    openSpeechModels: {},
-                    showMainWindow: {}
-                )
-            )
             host.reload(servers: model.settings.mcpServers)
         }
         .onChange(of: model.settings.mcpServers) { _, servers in
--- a/Sources/Nativ/Features/VoiceCapture/AudioView.swift
+++ b/Sources/Nativ/Features/VoiceCapture/AudioView.swift
@@ -894,7 +894,7 @@ struct AudioView: View {
         case .preparing:
             "Preparing \(captureLibrary.activeKind?.title.lowercased() ?? "recording")…"
         case .recording:
-            "Recording \(captureLibrary.activeKind?.title.lowercased() ?? "audio")"
+            "Recording"
         case .processing:
             "Saving and transcribing…"
         }


Blaizzy (commented):
Thank you, this mostly LGTM! Just have one question about the reported root cause: since `AppDelegate` is created through `@NSApplicationDelegateAdaptor` and owns `NativExtensionManager`, which in turn retains the host context/configuration provider, I would normally expect the delegate to remain alive for the app lifetime and the provider not to outlive it.

Could you share evidence that the captured `AppDelegate` is actually deallocated during reproduction—for example, a `deinit` trace or object-identity check—and clarify what still owns the extension manager/provider at that point? Capturing only `NativModel` looks safe, but without that lifecycle evidence I’m concerned this may be masking a different ownership issue.

Yavaren:
@Blaizzy of course, thanks. I will do everything right after I publish a UI fix PR.

Yavaren:
@Blaizzy `AppDelegate` was not deallocated during reproduction. The same `NativExtensionManager` instance remained alive throughout, and no `AppDelegate.deinit` trace occurred.

The diagnostics showed that `ExtensionsHubView` calls `manager.launch(...)` again with a placeholder transcription provider that returns `nil`. This replaces the valid host context originally installed by `AppDelegate`:

```text
[AudioLifecycle] manager.launch source=AppDelegate manager=ObjectIdentifier(0x0000000afe180a00)
[AudioLifecycle] NativExtensionManager.launch manager=ObjectIdentifier(0x0000000afe180a00) replacingHostContext=false
[AudioLifecycle] manager.launch source=ExtensionsHub manager=ObjectIdentifier(0x0000000afe180a00) placeholderProvider=true
[AudioLifecycle] NativExtensionManager.launch manager=ObjectIdentifier(0x0000000afe180a00) replacingHostContext=true
[AudioLifecycle] transcribe providerProperty=set invoking=true
[AudioLifecycle] provider source=ExtensionsHub result=nil
[AudioLifecycle] transcribe providerResult=nil

lucasnewman (approved):

lucasnewman:
The `AppDelegate` is effectively a singleton, so will never hit deinit. This looks good to me!

PR #330, opened 2026-08-19
Import an audio file into the audio library

Lazarus-931:
Adds an Import file action to the Recordings panel that copies or transcodes a chosen audio file into the library and runs the existing transcribe and summarize pipeline, redesigned from current main as a replacement for #253.


Lazarus-931:
superseded by #331

PR #331, opened 2026-08-20
Add audio upload to the library

Blaizzy:
## Summary
- add an Upload audio button to the Audio Library recordings panel
- import, transcribe, and automatically summarize selected audio files
- date uploaded items by import time so they appear at the top of the library
- migrate existing imported items without using `updatedAt` for ordering

## Testing
- focused AudioAnalyticsStore import-date regression tests
- unsigned Xcode build, recursive Apple Development signing, signature verification, and app launch

## Screenshot

Audio library upload UI

### DIFF
--- a/Sources/Nativ/Features/VoiceCapture/AudioAnalyticsStore.swift
+++ b/Sources/Nativ/Features/VoiceCapture/AudioAnalyticsStore.swift
@@ -239,7 +239,8 @@ final class AudioAnalyticsStore: ObservableObject {
         recordingURL: URL,
         kind: AudioRecordKind,
         title: String,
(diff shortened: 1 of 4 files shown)


Lazarus-931 (approved):
lgtm, but for the timestamp for file uploads, if timezones change after upload, it would incorrectly show the same time but in the new timezone.

like if i upload at 9am est, and move to pst(which is 3 hrs behind), it wont display 6am pst as the upload time

lucasnewman:
Can we use a different verb than "Upload"? I think "Add Recording" would be better and more clear that it's not leaving the device, and is consistent in the use of Audio vs Recording.

PR #332, opened 2026-08-20
Working draft: agent access to Nativ over MCP

Lazarus-931:
Working draft, not for review yet. Lets coding agents list and run Nativ's tools without opening a chat session.

Nativ listens on this Mac only. Agents on the same machine connect directly; a cloud agent needs the user to forward a public address to the second port themselves, and only gets a read-only subset of tools.


Blaizzy:
Thanks @Lazarus-931, let's park this for a future release.

PR #337, opened 2026-08-20
Add bulk selection across several features in nativ

Lazarus-931:
Multi-select was missing for multiple components of nativ such as:

1. Model deletion 
2. Audio Library (recent dictations and recordings0
3. Scheduled section
4. Extensions(Skills, mcp, tools and extensions)

Screenshot 2026-08-20 at 10 18 15 PM

Screenshot 2026-08-20 at 10 19 54 PM

Screenshot 2026-08-20 at 10 19 30 PM


lucasnewman:
@Lazarus-931 Sorry, lost track of this one. I think targeting the installed models page for the first rev would be ideal -- can we do that? I also think the wording and placement of the editing mode button could be improved a bit to make it more obvious about what it does.

Lazarus-931:
> @Lazarus-931 Sorry, lost track of this one. I think targeting the installed models page for the first rev would be ideal -- can we do that? I also think the wording and placement of the editing mode button could be improved a bit to make it more obvious about what it does.

that makes sense, i'll break these down independently

lucasnewman:
This is covered in the other merged bulk selection PRs now, closing.

PR #341, opened 2026-08-20
Reconcile routine launchctl jobs off the main thread to avoid blocking app startup

lucasnewman:
This was calling out to `launchctl` and waiting for it to exit on the main thread even though there was nothing to wait for. It also avoids any disk I/O if the routine is already up to date on disk.


Lazarus-931 (approved):
lgtm!

PR #342, opened 2026-08-20
Don't block the main thread when loading the stats database

lucasnewman:
Load them as empty first and then load the database async so we don't block app launch.


Lazarus-931 (approved):
👍🏽

PR #343, opened 2026-08-20
Check for software updates only after app launch is complete

lucasnewman:
This prevents blocking app launch on the SU network calls. I also opted in automatic updates by default (the modern standard) to avoid the annoying dialog that otherwise runs on first app launch.


Lazarus-931 (approved):
nice!

PR #346, opened 2026-08-21
Select drafter models in the chat picker to enable speculative decoding

BenedictKing:
## Problem

Drafter snapshots such as `mlx-community/Qwen3.8-27B-MTP-8bit` carry text-generation tags, so they showed up in chat model pickers. Selecting one loaded the MTP draft model as the main chat model, and the server crashed handling requests:

```
'Qwen3_5MTPDraftModel' object has no attribute 'language_model'
```

(raised by `BatchGenerator(self.model.language_model, ...)` in mlx-vlm's server generation path — a drafter can only run as `--draft-model` of a compatible target, never standalone).

But users who select an MTP model in the picker actually want the `--model Qwen3.8-27B-8bit --draft-model Qwen3.8-27B-MTP-8bit` speculative-decoding combo, not to load the drafter itself.

## What changed

**Drafters can no longer be loaded as chat models.**
- All language-model picker entry points (chat composer, scheduled tasks, model preloading, integrations) now filter with `isEligibleForLanguageModelPicker`, which excludes `drafter`/`reranking` capabilities.
- `NativModel.startServer` skips preloading a known drafter left over in stored settings; the server's `/models/load` rejects drafters as the main model with a 400.

**Selecting a drafter in the chat picker now enables speculative decoding on its target.**
- The chat composer model menu lists drafters in their own "Drafters — speculative decoding" section with a kind badge (e.g. `Qwen3.8-27B-MTP-8bit (MTP)`).
- Selecting one resolves its compatible chat model via `DrafterTargetResolver`: keep the current model when compatible (hidden-size match, missing metadata counts as compatible), then the name-derived pair (`X-27B-MTP-8bit` → `X-27B-8bit`; mtp/eagle3/eagle/dflash markers), then the first compatible chat model. Entries with no compatible target are annotated and inert.
- The selection writes `speculativeDecodingEnabled`/`draftModelID`/`draftKind=auto`, persists it into the target model's per-model profile, and restarts or switches the server with the drafter attached.

## Testing

- New `DrafterTargetResolverTests` (target resolution: current model, name derivation, hidden-size fallback, no-compatible → nil) and an MTP snapshot scanning test (`qwen3_5_mtp` classified as ineligible drafter).
- Full `NativTests` suite passes locally (443 tests, 0 failures).
- Verified end-to-end on device: selecting `Qwen3.8-27B-MTP-8bit (MTP)` in the composer menu restarts the server with `Drafter ready; speculative decoding enabled.` and responses report `draft_kind: mtp` in timings.

### DIFF
--- a/Docs/features/models.md
+++ b/Docs/features/models.md
@@ -29,9 +29,10 @@ Each model carries a set of capabilities (`LocalModelCapability`):
 | `tools` | Supports tool calling. |
 | `drafter` | Usable as a speculative-decoding draft model. |
 
-Capability determines where a model is offered: only `text` (and non-image `vision`) models
-appear in the conversation model picker; image, speech, and embedding models are selected in
-their own slots.
+Capability determines where a model is offered: chat model pickers use
+`isEligibleForLanguageModelPicker` — `text` (and non-image `vision`) models qualify, while
+`drafter` and `reranking` models are always excluded. Image, speech, and embedding models are
+selected in their own slots.
 
 ## Downloading
 
--- a/Docs/features/models.md
+++ b/Docs/features/models.md
@@ -75,3 +75,12 @@ A draft model can accelerate a larger target model. Enable `speculativeDecodingE
 `draftModelID`; the draft must be capability-compatible with the target. Draft kind and block
 size are configurable (`draftKind`, `draftBlockSize`). Only `drafter`-eligible models are offered
 as drafts, and a mismatched hidden size is flagged as an incompatible target.
+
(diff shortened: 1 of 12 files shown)


Lazarus-931:
hey @BenedictKing, we're currently working on spec decode in mlx-vlm, some issues remain in the engine. a follow up pr will be put up for a formal/official method of using models with their spec decode counterpart

lucasnewman:
@BenedictKing Thanks for the contribution! We added a version of this for our next update that should make it easier to select draft models -- give it a shot and let us know if you think it needs any additional improvements.

PR #347, opened 2026-08-21
Show drafter models as drafters instead of Not supported

BenedictKing:
## Problem

MTP/EAGLE/DFlash drafter models (e.g. `mlx-community/Qwen3.8-27B-MTP-8bit`) aren't eligible for the language picker — they're attached to a compatible chat model via Chat settings → "Enable drafter" for speculative decoding instead. Their row in Models should say so, but the current treatment depends on the drafter's `model_type`:

- Drafters whose `model_type` isn't language-mapped fall into the `preloadSlots.isEmpty` branch and show an orange "Not supported" pill plus an "Unavailable" button.
- `qwen3_5_mtp` resolves as a language model type, so MTP checkpoints got a Language preload slot and a working preload button — bypassing the language picker's eligibility filter and offering to "preload" a drafter that can't run standalone.

## Fix

Render drafter-specific UI instead:

- `preloadSlots(for:)` now returns empty for models with the `.drafter` capability — drafters never take a preload slot, so the row treatment below applies to every drafter kind
- Title pill reads "MTP drafter" (falls back to "Drafter") in purple instead of "Not supported"
- Action button reads "Drafter" with the `hare` icon instead of "Unavailable"
- The info alert explains the model is a speculative-decoding drafter and points to Chat settings → "Enable drafter"
- Hover help text updated accordingly

Verified with `make xcode-build` (Debug, BUILD SUCCEEDED) and by checking the row, pill, and info alert render correctly with `mlx-community/Qwen3.8-27B-MTP-8bit` in the Hugging Face cache.

### DIFF
--- a/Sources/Nativ/Features/Models/ModelsView.swift
+++ b/Sources/Nativ/Features/Models/ModelsView.swift
@@ -801,6 +801,12 @@ struct ModelsView: View {
     }
 
     private func preloadSlots(for localModel: LocalModel) -> [ModelPreloadSlot] {
+        // Drafter models attach to a chat model via Chat settings and never
+        // take a preload slot, even when their model_type resolves as a
+        // language model (e.g. qwen3_5_mtp).
+        guard !localModel.capabilities.contains(.drafter) else {
+            return []
+        }
         var slots: [ModelPreloadSlot] = []
         if localModel.capabilities.contains(.text)
             && !localModel.capabilities.contains(.reranking) {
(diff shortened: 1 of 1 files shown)


Lazarus-931:
I think this is an ok temporary solution until drafts are supported, can you show a screenshot of what it looks like?

BenedictKing:
Here's what it looks like with an MTP drafter (`mlx-community/Qwen3.8-27B-MTP-8bit`) in the Hugging Face cache.

The Models list shows a purple **MTP drafter** pill on the row instead of the old "Not supported", with regular supported models below it for comparison:

Models list with the MTP drafter pill

Clicking the row's **Drafter** button explains what the model is and where to enable it:

Drafter model info alert

PR #353, opened 2026-08-21
Rework control panel for smooth animations and consistent layout

lucasnewman:
This is a rework of the main view to have proper window controls, smooth animation for sidebar panels, and consistent materials. I also aligned all headers for the detail view to exactly match the sidebar so we have visual consistency throughout the app.

It preserves the fullscreen behavior with sharp corners and edge-to-edge panels without the need for a custom window controls implementation and any kind of internal view modification (which can easily break between OS releases), and gets us a net reduction of 1k+ LoC across the control panel implementation.

https://github.com/user-attachments/assets/62cc43b1-889d-42e6-ad33-a63cea50d87e


Lazarus-931 (approved):
lgtm!

PR #356, opened 2026-08-22
Migrate SystemMonitorStore to @Observable

Lazarus-931:
Another polish, the system monitor updated views it didn't need to, this makes observation proerty-level instead. 11 values are published every sec, under `ObservableObject`

With `@Observable`, views re-render only when a property they actually read changes

No API or behavior change: same `start/stop/refresh/collectSnapshot` surface, same sampling cadence.

### DIFF
--- a/Sources/Nativ/Features/SystemMonitor/SystemMonitorStore.swift
+++ b/Sources/Nativ/Features/SystemMonitor/SystemMonitorStore.swift
@@ -190,10 +190,10 @@ final class SystemMonitorStore {
     private let collector = SystemMetricsCollector()
     private let displayFPSSampler = SystemDisplayFPSSampler()
     private let aneSampler = SystemANEUtilizationSampler()
-    nonisolated(unsafe) private var samplingTask: Task<Void, Never>?
+    private var samplingTask: Task<Void, Never>?
     private let historyLimit = 300
 
-    deinit {
+    isolated deinit {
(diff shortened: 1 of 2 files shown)


Sources/Nativ/Features/SystemMonitor/SystemMonitorStore.swift
```diff
@@ -172,24 +172,25 @@ struct SystemMonitorSnapshot: Equatable, Sendable {
...
+    private(set) var diskWriteHistory: [SystemHistorySample] = []
+    private(set) var temperatureHistory: [SystemHistorySample] = []
+    private(set) var powerHistory: [SystemHistorySample] = []
+    private(set) var isSampling = false
 
     private let collector = SystemMetricsCollector()
     private let displayFPSSampler = SystemDisplayFPSSampler()
     private let aneSampler = SystemANEUtilizationSampler()
-    private var samplingTask: Task<Void, Never>?
+    nonisolated(unsafe) private var samplingTask: Task<Void, Never>?
```
lucasnewman: Can this be marked `@ObservationIgnored` instead?

Lazarus-931: I was having build issues with `deinit` but after re-visiting, I did manage to figure out why, i'll fix this

lucasnewman (approved):

PR #357, opened 2026-08-22
Move server logs into a dedicated observable store

Lazarus-931:
Incoming server logs re-rendered unrelated screens. Now they update only the developer log views.

Accumulation now lives in a dedicated `@MainActor @Observable ServerLogStore`; only the developer log panel and issue reports read it, so log traffic invalidates just those views.


lucasnewman (approved):
👌

PR #359, opened 2026-08-22
Migrate NativModel to @Observable

Lazarus-931:
The app-wide model invalidated every screen on every change. Now screens update only when values they read change. 

This cuts down cpu/gpu usage, since pages were doing un-necessary redraws


Sources/Nativ/Features/ControlPanel/ControlPanelState.swift
```diff
@@ -71,44 +73,32 @@ final class ControlPanelChromeState: ObservableObject {
...
+    private func observeModel() {
+        withObservationTracking { [weak self] in
+            self?.captureModelSnapshot()
+        } onChange: { [weak self] in
+            Task { @MainActor [weak self] in self?.observeModel() }
+        }
+    }
+
+    private func captureModelSnapshot() {
+        update { $0.isRunning = model.isRunning }
```
lucasnewman: Can we use a single `update { ... }` block here to coalesce changes? I'm not sure if it will fire onChange for each block so it may not be a real problem, but it seems like best practice to have a unified update in any case.

Sources/Nativ/Features/Models/ModelsView.swift
```diff
@@ -91,43 +92,36 @@ private final class ModelsNativState: ObservableObject {
...
+        _ keyPath: ReferenceWritableKeyPath<ModelsNativState, T>,
+        to value: T
+    ) {
+        if self[keyPath: keyPath] != value {
+            self[keyPath: keyPath] = value
+        }
+    }
+
+    private func captureModelSnapshot() {
+        update(\.settings, to: model.settings)
```
lucasnewman: I think the pattern in the ControlPanelState is better so we can coalesce the updates here as well.

lucasnewman (changes requested):
@Lazarus-931 Overall this is great, see my comments for a minor tweak we should make.

lucasnewman:
@Lazarus-931 Looks great once the conflicts are resolved!

lucasnewman (approved):

PR #361, opened 2026-08-22
Render the chat transcript lazily

Lazarus-931:
Transcript rows were all built upfront even off-screen; LazyVStack now builds them on demand, keeping long chats scrolling smoothly with identical behavior.


lucasnewman:
This doesn't work because the estimated layout height is incorrect and will break the progress bar and causing rendering artifacts. We would need accurate estimates or (more likely) cached layout heights for this to work properly.

PR #362, opened 2026-08-22
Store the Hugging Face token in the Keychain

Lazarus-931:
hf tokens sat unencrypted inside Settings.plist. It now lives in the Keychain and migrates automatically on launch.


lucasnewman (approved):
Thanks!

PR #363, opened 2026-08-22
Secure local server access

Lazarus-931:
Restrict the bundled server to loopback addresses and require a token for every supported request.

Generate a token automatically, store it in Keychain, and migrate legacy settings without exposing secrets.

Update integration profiles to use the generated token instead of the previous hardcoded fallback value.

Add regression coverage for startup policy, middleware behavior, credential migration, packaging, and integration configuration correctness.


lucasnewman:
What's the intention here? Wouldn't this prevent using the server on a LAN from another application?

Issue #325, opened 2026-08-18
Harden Core Audio tap isolation with regression coverage and coalesced UI updates


Blaizzy:
## Context

PR #320 fixed #319 by declaring the Core Audio tap callbacks as `@Sendable`. This prevents the callbacks from inheriting `@MainActor` isolation under Swift 6.3 when AVAudioEngine invokes them on its real-time audio queue.

The correctness fix is appropriate, but two follow-ups would make this path more robust:

1. Add regression coverage for off-main Core Audio callback execution.
2. Avoid creating a new main-actor `Task` for every audio buffer when only meter/UI updates need to cross actors.

## Scope

- Refactor the tap-processing boundary into a testable `@Sendable` callback or helper.
- Add a regression test that invokes tap processing from a non-main executor and verifies:
  - no actor-isolation assertion/trap;
  - audio calculation/writing remains off the main actor;
  - observable meter state is updated only on the main actor.
- Exercise repeated Audio extension enable/disable followed by opening the Audio page, matching the #319 reproduction where practical.
- Evaluate coalescing or throttling main-actor meter updates so the tap does not allocate one `Task` per 1,024-frame buffer.
- Cover both `AudioInputLevelMonitor` and `VoiceAudioRecorder`.
- Preserve current recording correctness, the 15 Hz UI update cap, and real-time callback safety.

## Acceptance criteria

- Automated coverage fails for the pre-#320 actor-isolated callback pattern and passes with the corrected boundary.
- Repeatedly toggling the Audio extension and opening Audio does not crash.
- Metering and recording behavior remain unchanged from the user's perspective.
- Main-actor handoffs are bounded/coalesced rather than scheduled for every incoming audio buffer.
- The project builds cleanly with Swift 6.3 concurrency checking enabled.

Related: #319, #320

cc @Yavaren

lucasnewman:
Resolved in https://github.com/Blaizzy/nativ/pull/364

PR #364, opened 2026-08-22
Move audio meter capture off the main thread

lucasnewman:
This adds a publisher which computes the meter levels off the main thread and delivers them to the main thread on a 15Hz timer to coalesce UI updates.

https://github.com/Blaizzy/nativ/issues/325


Lazarus-931 (approved):
lgtm!

PR #365, opened 2026-08-22
fix: reset prefix cache across model eviction and reload

### LINKED ISSUE #272: v0.3.0: chat completions hang after model eviction with prefix caching enabled — worker spins at ~94% CPU, /health unresponsive
## Summary

On v0.3.0, after a large streaming `/v1/chat/completions` request completes, the model is evicted (existing load-on-demand behavior). A **second** large request sent within \~30s never completes: the `mlx-vlm-server` worker process spins at \~90–94% CPU with near-zero RSS, `/health` stops responding, and every subsequent request times out. Only killing the app recovers. With prefix caching **disabled**, the same rapid-fire pattern works reliably (3 consecutive full-size agent loops today, all completed).

## Environment

* macOS 26.5.2, Apple M4, 24 GB unified memory
* Nativ **v0.3.0** (2026-08-10 release), server on `127.0.0.1:8080`
* Model: `mlx-community/Qwen3-14B-4bit`
* Settings at time of hang (Settings.plist):
  * `kvQuantizationEnabled: true` + `turboQuantEnabled: true` → server flags `--kv-bits 8 --kv-quant-scheme turboquant --kv-group-size 64 --quantized-kv-start 0`
  * `prefixCachingEnabled: true` (→ `apc_enabled: true`)
  * `maxKVSize: 32768`

## Repro

1. Launch Nativ, select Qwen3-14B-4bit.
2. Send a **large streaming** chat completion (\~16,300-token system prompt, `max_tokens: 8192`, `stream: true`):

   ```bash
   curl -s -N http://127.0.0.1:8080/v1/chat/completions \
     -H "Authorization: Bearer <key>" -H "Content-Type: application/json" \
     -d '{"model":"mlx-community/Qwen3-14B-4bit","stream":true,"max_tokens":8192,
          "messages":[{"role":"system","content":"<~16k-token prompt>"},
                      {"role":"user","content":"hi"}]}'
   ```

   → Completes normally. Telemetry: 16,321 prompt tokens, prefill 45.1 tok/s, total 369.9s.
3. Within \~15–30s of completion, send a second request with the same system prompt (different user tail).<br>→ **Never completes.** The worker's RSS drops from \~7.3 GB to \~0.02 GB (model evicted), the process spins at \~90–94% CPU, `/health` stops responding (5s timeout), and any further chat request times out (12s+). App must be killed to recover.

## Expected vs actual

* **Expected:** the second request triggers a model reload (as it does with prefix caching off) and completes.
* **Actual:** worker deadlock/spin after eviction; server unresponsive until app kill.

## Additional observations

* Eviction is visible in RSS: after request 1 completes the worker holds \~7.3 GB; during the hang it's \~0.02 GB — yet `/health` continues to report `loaded_model` (existing quirk: health reports configured, not resident).
* With `prefixCachingEnabled: false` (same KV-quant settings), consecutive full-size requests work. Today's back-to-back agent loops, all exit 0:
  * 08:54 → 217.5s (prefill 79.0 tok/s) — apc OFF
  * 09:09 → 369.9s (prefill 45.1 tok/s) — apc ON, first large request after enabling (worked)
  * 09:23 → 279s — apc OFF
* The hang appeared on the **second** large request with apc ON (09:16:20). The failed request never appears in `request_events` (Analytics.sqlite3).
* A small probe request fired \~5 min into the hang also timed out — though the worker was already spinning at 94% CPU with 0.02 GB RSS before the probe was sent.
* A single small streaming request with apc ON works fine (2.1s) — the issue is specific to the eviction → reload path.

## Workaround

Set `prefixCachingEnabled: false` in `~/Library/Application Support/Nativ/Settings.plist`, restart the app. Consecutive large requests then complete reliably. (We're keeping prefix caching off until this is resolved.)

## Related

* Blaizzy/nativ#245 (feature request: model release and startup auto-load controls) — the eviction/reload cycle is the underlying behavior; a keep-loaded option would sidestep this class of bug.
* Blaizzy/nativ#255 (empty streaming responses) — fixed in v0.3.0, verified working here (SSE now flows correctly for a single request with apc ON).

## Impact

API-driven / agent workloads that issue consecutive large requests (tool-calling agents, batch jobs) can deadlock the server after the first completion, requiring an app restart. This makes server mode unreliable for repeated calls even though single requests (streaming or not) work fine.

mvanhorn:
Extend the existing runtime patch installation in `nativ_server.py` so every upstream path that releases the resident text model also resets the associated APC manager through its supported cache-reset behavior before the model cache becomes reusable. Nativ's bundled `mlx-vlm-server` can retain automatic-prefix-cache state after an on-demand model is evicted.

**Tests**

- With APC enabled: simulate a completed request, model eviction, and a second
  request. Verify eviction clears the old APC contents and that reload
  initializes against a clean cache rather than retaining references to the
  released model.
- With APC disabled: run the same eviction/reload lifecycle and verify the
  guard is a no-op and existing model reload behavior is unchanged.

Fixes #272

### DIFF
--- a/PythonDistribution/Overlay/nativ_server.py
+++ b/PythonDistribution/Overlay/nativ_server.py
@@ -10,6 +10,7 @@ import uuid
 from contextvars import ContextVar
 from dataclasses import dataclass
 from datetime import datetime
+from importlib import import_module
 from threading import Lock
 from types import SimpleNamespace
 from typing import Any
@@ -25,6 +26,7 @@ import mlx_vlm.server.generation as base_generation
 import mlx_vlm.server.openai as base_openai
 
 
+base_app = import_module("mlx_vlm.server.app")
 BACKEND_NAME = f"mlx_vlm/{base.__version__}"
 TRACKED_PATHS = {
     "/chat/completions",
@@ -134,6 +136,38 @@ def install_model_load_progress() -> None:
     base_generation.load = load_with_progress
     base_generation.ResponseGenerator._initialize_model = initialize_model_with_progress
     base_generation._nativ_model_load_progress_installed = True
+
+
+def install_model_lifecycle_cache_reset() -> None:
+    """Clear text APC state before upstream releases its resident model."""
+    if getattr(base_app, "_nativ_model_lifecycle_cache_reset_installed", False):
+        return
+
+    original_unload_model_cache_group = getattr(
+        base_app,
+        "_unload_model_cache_group",
+        None,
+    )
+    if original_unload_model_cache_group is None:
+        return
+
+    def unload_model_cache_group_with_apc_reset(cache_group: str) -> bool:
+        if cache_group == "text_generation":
+            registry = base_app._model_cache_registry()
+            cache = registry.for_kind(cache_group)
+            apc_manager = cache.get("apc_manager") if cache else None
+            if apc_manager is not None:
+                apc_manager.clear()
+
+        return original_unload_model_cache_group(cache_group)
+
+    base_app._unload_model_cache_group = unload_model_cache_group_with_apc_reset
+    if (
+        getattr(base, "_unload_model_cache_group", None)
+        is original_unload_model_cache_group
+    ):
+        base._unload_model_cache_group = unload_model_cache_group_with_apc_reset
+    base_app._nativ_model_lifecycle_cache_reset_installed = True
 
 
 class MetricsAccessLogFilter(logging.Filter):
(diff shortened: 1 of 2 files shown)


Lazarus-931:
hi @mvanhorn , thx for the pr!

may i ask why this issue is addressed in mlx-vlm directly?

lucasnewman:
@mvanhorn Thanks for reporting this! I've got a minimal fix for our upstream project that should resolve this for our next release: https://github.com/Blaizzy/mlx-vlm/pull/2010

lucasnewman:
Resolved in https://github.com/Blaizzy/mlx-vlm/pull/2010 for the next release.

PR #368, opened 2026-08-23
Bump version to 0.3.4

lucasnewman:
Also added a script to make it easier to bump the version. It does patch by default but also takes an explicit version as an optional param.


Lazarus-931 (approved):
👍🏽

PR #371, opened 2026-08-24
Decouple scheduled tasks from chats

Blaizzy:
## Summary

- make scheduled tasks independent, durable objects
- create a unique, visible chat for every scheduled execution
- keep the task and run record when a linked chat is deleted
- offer two explicit task-deletion choices: keep linked chats as regular chats, or delete them too
- stop creating or restoring permanent source chats for scheduled tasks
- cancel queued or active execution when a task is deleted and prevent late completion from recreating deleted state

## Previous behavior

Scheduled tasks used a chat-first model: creating a task created a permanent source chat, every execution appended to that same chat, and the app restored the source chat on launch. Deleting the linked chat therefore cascaded into deleting the scheduled task, its run records, and its associated chats.

That ownership model was internally consistent when the chat represented the task itself, but it no longer matched the Scheduled UI, where tasks appear to be independent, durable objects. Deleting what looked like a conversation could unexpectedly disable the automation and erase all run history. Reusing one transcript also made individual executions harder to inspect, delete, and diagnose independently.

```
Former: Chat owns Task → Task owns Runs
New:    Task owns Runs → Each Run links to a Chat
```

### Delete task choices

Delete scheduled task dialog showing Keep Chats and Delete Chats options

### Delete Task, Keep Chats

The task and run history are removed. Its linked chats remain as regular conversations.

Scheduled tasks after keeping and detaching linked chats

### Delete Task and Chats

The task, run history, and linked chats are removed. Only previously detached chats remain.

Scheduled tasks after deleting the task and all linked chats

## Testing

- `xcodebuild test -project Nativ.xcodeproj -scheme Nativ -destination 'platform=macOS' -derivedDataPath build/NativDevelopmentDerivedData -only-testing:NativTests/ScheduledTaskChatLinkerTests -only-testing:NativTests/ChatViewModelTests CODE_SIGNING_ALLOWED=NO`
- 6 scheduled-task lifecycle tests passed
- 4 ChatViewModel tests passed
- app target compiled successfully as part of the test build
- manually verified both task-deletion outcomes in the built macOS app using an isolated profile
- GitHub `build` check passed


Lazarus-931 (approved):

lucasnewman (approved):
LGTM

PR #372, opened 2026-08-24
Add multi-window support

### LINKED ISSUE #349: Support multiple windows
It would be really useful to be able to open a view in a separate windows. I'd like to be able to monitor progress in the chat view, while at the same time monitor system specs in the System view and/or server logs in the Dev view.

Probably a few ways to approach this, but I could envision the app working as it does currently by default. Then, while in a given view (e.g., Dev is the current view), pop that view into a new window.

p.s. Thanks for making this app. It's one of the coolest tools I've installed in a while.

Yavaren:
## Summary

Closes #349.

This PR adds support for running multiple independent Nativ windows.

Each window has its own:

- Navigation
- Selected chat
- Draft and attachments
- Chat and image-generation workflow
- Sidebar and presentation state

Application-wide resources remain shared:

- Nativ model and inference server
- Loaded models
- MCP connections
- System monitoring
- Settings and downloads

## Window behavior

- `Command-Shift-N` opens a new window.
- A new window can be opened through file / new window

## Concurrent activity

- Different windows can use different chats independently.
- The same chat or image session cannot be modified from another window while it is generating.
- In the second window, the unavailable Send button appears gray.
- Model selection is disabled in every window while generation is active.
- Model menus remain openable and explain why the model cannot be changed.
- Persisted chat and image-session changes are shared between windows.

## System monitoring

System monitoring remains application-scoped. Multiple windows can observe it without starting duplicate samplers.

## Testing

- 457 tests passed.
- Static analysis passed.
- Multiple-window opening and closing was manually verified.
- Only one shared application process and shared resource set are used.

Screenshot 2026-08-24 at 5 35 45 PM
Screenshot 2026-08-24 at 5 35 31 PM
Screenshot 2026-08-24 at 5 44 07 PM
Screenshot 2026-08-24 at 5 43 59 PM


Sources/Nativ/Features/ImageGeneration/ImageGenerationViewModel.swift
```diff
@@ -705,6 +771,9 @@ final class ImageGenerationViewModel: ObservableObject {
     }
 
     func removeOutput(sessionID: UUID, turnID: UUID, outputID: UUID) {
+        guard canModifySession(sessionID) else {
```
Lazarus-931, line 775: Can this return a success value and let `ArtifactStore` delete the file only on success? Right now this skips the session update, but the file still gets deleted and leaves a broken attachment in the history.

Lazarus-931 (commented):
one question, but looks great!

Sources/Nativ/Features/ImageGeneration/ImageGenerationViewModel.swift
Yavaren: @Lazarus-931 yes you are right, thank you. Working on a fix for this.

Yavaren: Fixed, lmk what you think!

lucasnewman:
@Yavaren This PR mixes a lot of things into a single change -- can we stack the changes to make it more reviewable? There's a bunch of pre-work that can be done before enabling multi-window support itself:

1) The ref counted system-monitor observers
2) Dependency injection / ownership changes
3) Persistence and cross-instance sync
4) Model switching policies
5) Transactional artifact deletion
6) Command routing

Once we have those in place we can then flip the switch over to use WindowGroup. Sound good?

Yavaren:
> @Yavaren This PR mixes a lot of things into a single change -- can we stack the changes to make it more reviewable? There's a bunch of pre-work that can be done before enabling multi-window support itself:
> 
> 1. The ref counted system-monitor observers
> 2. Dependency injection / ownership changes
> 3. Persistence and cross-instance sync
> 4. Model switching policies
> 5. Transactional artifact deletion
> 6. Command routing
> 
> Once we have those in place we can then flip the switch over to use WindowGroup. Sound good?

Agreed, thanks. I will start the work in several hours.

PR #373, opened 2026-08-24
Fix scrolling performance for long user prompts

lucasnewman:
Use a native SwiftUI text view instead of the bridged AppKit view for user prompts, which has equivalent behavior but significantly improved layout performance.


Lazarus-931 (approved):
👍🏽

PR #375, opened 2026-08-24
Improving file processing

Lazarus-931:
This is substantially better than before at properly allowing models to work with files. I set it up so that it's extensible per file type and consists of changes that can grow with the file types and overall how user/agent works in nativ


Lazarus-931:
### Measured impact: before → after

Before this PR, document context was the first 24,000 characters; at 600,000 characters, retrieval recall was **0%**. After this PR, Nativ caches the full extraction and selects query-relevant sections at request time: recall at 600,000 characters reached **95%** in the headline comparison and **90% overall** in the broader retrieval matrix. Across 10 local models, answer accuracy was **93% when the evidence reached context and 0% when retrieval dropped it**—no model recovered a retrieval miss.

This was measured with a standalone SwiftPM harness vendoring the PR sources verbatim and passing the PR's 18 relevant tests: 106 real files plus 75 generated documents (2k–600k characters), 1,717 retrieval cells, and 2,376 model requests across 10 models. Token counts used Nativ's bundled tokenizers and were cross-checked against server-reported `prompt_tokens`. The decision signal is clear: improve retrieval before upgrading models; token-aware budgets and dropped-document visibility remain the next separate decisions.

lucasnewman (approved):
Nice!

PR #389, opened 2026-08-25
Add an optional tool to read files safely from a user-selected folder

lucasnewman:
This adds a new tool `read_file` which allows reading files (text and PDF only at the moment) from a user-selected folder, automatically limiting the content size to avoid exploding the context window.

I've set it up to be intentionally conservative for now, so It's off by default and can only be enabled by selecting a folder in which it's allowed to read from. It will automatically redact any secret values for common secrets.

Screenshot 2026-08-25 at 12 27 40 PM
Screenshot 2026-08-25 at 12 27 34 PM
Screenshot 2026-08-25 at 12 28 05 PM


Lazarus-931 (approved):
conflicts, awesome otherwise!

Sources/Nativ/Features/Chat/Tools/ChatReadFileTool.swift
```diff
@@ -0,0 +1,709 @@
+import Foundation
+import NativServerKit
+
+enum ChatReadFileToolRegistry {
+    static let toolName = "read_file"
+    static let defaultOffset = 1
+    static let defaultLimit = 2_000
+    static let maximumLimit = 2_000
+    static let defaultMaximumResultCharacters = 100_000
```
Lazarus-931: would this overflow small context for smaller models?

in #375, the doc context is fit to the model's token budget

PR #392, opened 2026-08-26
Document Homebrew installation path

### LINKED ISSUE #57: Allow installation as Homebrew Cask
Currently the README recommends installation via direct download of the DMG, or building from source: https://github.com/Blaizzy/nativ/blob/main/README.md?plain=1#L73-L96.

I'm probably not alone in preferring to install as much software as possible via Homebrew, which can handle GUI applications via its Cask feature: https://github.com/Homebrew/homebrew-cask.

Main ask of this issue is to understand if this is already planned, and if so to offer my help in getting its formula accepted upstream.

jacquerie:
Closes https://github.com/Blaizzy/nativ/issues/57

PR https://github.com/Homebrew/homebrew-cask/pull/283062 got merged, so this PR does the final step of documenting this alternative installation path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,11 @@ To build from source, you will also need:
 
 ### Download a release
 
-Download the latest DMG from [GitHub Releases](https://github.com/Blaizzy/nativ/releases/latest), drag **Nativ** to Applications, and launch it. Nativ uses Sparkle for subsequent in-app updates.
+Download the latest DMG from [GitHub Releases](https://github.com/Blaizzy/nativ/releases/latest), drag **Nativ** to Applications, and launch it. Alternatively, install it via Homebrew:
+```sh
+brew install --cask nativ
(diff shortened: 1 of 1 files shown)


lucasnewman:
@jacquerie Thank you!

abustany:
Hmm I just installed Nativ via Homebrew (used to run a manual install before), and I had to remove the `com.apple.quarantine` manually (macOS would not let me open the app otherwise). Not sure if something is funky with my setup or with the brew pipeline.

jacquerie:
I'm tempted to call it something funky with your setup, because the formula is as vanilla as it gets... That said, is this something you can reliably reproduce by uninstalling / installing, or just something you witnessed once? If it's the former, can you open an issue against `homebrew-cask` with all the diagnostic data they ask you to provide?

PR #251, opened 2026-08-10
Feature/custom kits

Lazarus-931:
Allows people to use custom kits, this pr does several things:

- Add user-created Kits composed from extensions, MCP servers, tools, and skills.
- Redesign Kit creation and component selection for a compact, tag-based workflow.
- Let tool-capable models discover and use both built-in and user-created Kits.
- Require confirmation before a model activates a Kit.
- Apply Kit settings immediately and wait for MCP connections so newly enabled MCP tools are available on the next model turn.

PR #395, opened 2026-08-26
Resolve Kit capabilities at routine runtime

Lazarus-931:
Resolves selected Kits against current settings while preserving stable identities and deduplicating shared runtime capabilities. Fails routines early with named unavailable components instead of silently executing partially configured Kit selections. Adds focused coverage for disabled skills, missing servers, unknown Kits, and shared capability deduplication behavior.

Mainly helps by preventing routines from silently running incomplete Kits.

This is pre-cleaning for an upcoming pr to supercede #251


lucasnewman (approved):
LGTM

PR #397, opened 2026-08-26
Make Kits scalable and accessible

Lazarus-931:
Makes Kit cards and management sheets adapt cleanly across window sizes and expanding catalogs naturally.

Replaces gesture-only interactions and fixed typography with explicit controls, semantic styles, and compact status indicators.

Removes redundant activation helpers while resolving each Kit snapshot once per rendered interface surface update.

Mainly helps by keeping Kits usable as their catalog and component counts grow.

This is pre-cleaning for an upcoming pr to supercede #251

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatCapabilitiesSheet.swift
+++ b/Sources/Nativ/Features/Chat/ChatCapabilitiesSheet.swift
@@ -283,27 +283,35 @@ struct ChatKitsPickerSheet: View {
     @Environment(\.dismiss) private var dismiss
 
     var body: some View {
+        let kits = NativKitCatalog.bundled.kits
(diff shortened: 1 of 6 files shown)


Lazarus-931:
simplified it so it's like:

image


image

image


Coloring is simplified with a half empty, empty and full circle also allow resizing the popup

lucasnewman:
@Lazarus-931 Can we call it "Enable All" even in the partial / missing case? I think "Enable missing" is somewhat confusing. Also, we should use title case (Enable All vs Enable all) for all buttons. I know we've been inconsistent thus far but it's the platform standard.

lucasnewman (approved):

PR #401, opened 2026-08-26
Standardize Product Terminology and Value Formatting

Lazarus-931:
Standardizes Nativ terminology and user-facing value formats for consistent labels, durations, rates, counts, and units.


lucasnewman (approved):
Nice, thanks!

PR #409, opened 2026-08-27
add model registry

Yavaren:
## Summary

This PR introduces a unified model capability registry shared by the bundled Python backend and the Swift runtime.

It replaces the image-specific manifest generation with a general capability manifest that records which model types are supported for language, vision, audio, image generation, image editing, embeddings, reranking, and speculative decoding.

The Swift runtime loads this generated information through `NativModelTypeRegistry`, providing one source of truth for model-type support.

## What changed

- Added a unified model-capability manifest generator.
- Added `NativModelTypeRegistry` to `NativServerKit`.
- Connected the registry to the existing server runtime.
- Replaced the previous image-only manifest generation.
- Updated the Xcode project to include the new registry source.

## Why

Model support was previously determined in several different places. That made it possible for discovery, installed-model handling, and the bundled backend to disagree about whether a model type was supported.

The registry provides a shared foundation for later PRs to make those decisions consistently.

## How it works

During the app build, the manifest generator examines the exact versions of the MLX packages bundled with Nativ. It identifies their available model loaders, capability declarations, and model-type aliases, then writes them into a versioned JSON manifest.

At runtime, Swift validates and loads that manifest. Other parts of Nativ can then determine whether the bundled runtime contains a candidate loader for a particular model type and which capabilities that loader provides.

A registry match means that Nativ ships a candidate loader for the model type. It does not guarantee that every individual repository using that type will load successfully.

The generator extends the same source-inspection approach previously used by the image-model manifest. It now covers additional model families and performs stricter validation. Because it depends on structures inside the bundled MLX packages, it should be reviewed whenever those dependencies are upgraded.

## Stack

1. **Model registry — this PR**
2. Model support detection
3. Active-download UI

This PR targets `main` and has no dependency on another PR in the stack.


lucasnewman (approved):
LGTM

PythonDistribution/Scripts/generate_model_capabilities_manifest.py
```diff
@@ -0,0 +1,365 @@
...
+    "language",
+    "speculative_drafters",
+    "image_generation",
+    "image_editing",
+    "speech_to_text",
+    "text_to_speech",
+    "embeddings",
+    "reranking",
+)
+PACKAGE_NAMES = ("mlx-lm", "mlx-vlm", "mlx-audio")
```
Blaizzy, line 24: Do we still use LM? 🤔

@lucasnewman @Lazarus-931

PR #410, opened 2026-08-27
detect model support

Yavaren:
## Summary

This PR connects the model capability registry to Hugging Face discovery and the local model library.

Models discovered on Hugging Face are classified as supported, unsupported, or unknown using their published configuration and the loaders bundled with Nativ.

Unsupported models remain visible and downloadable. Installed models also remain visible and can still be loaded manually, even when Nativ does not recognize a supported preload role.

## What changed

- Added a model-support classifier backed by `NativModelTypeRegistry`.
- Requested model configuration metadata from the Hugging Face API.
- Classified models as supported, unsupported, or unknown.
- Added an orange `Unsupported` badge to conclusive unsupported results in Discover.
- Kept unsupported models downloadable.
- Kept unrecognized Hugging Face cache entries visible on the Installed page.
- Allowed installed models without a recognized preload role to be loaded as language models.
- Used model configuration metadata to improve provider detection.
- Removed duplicated hard-coded model-support checks from `NativModel`.

## Why

The model registry introduced in the previous PR describes the loaders bundled with Nativ, but it does not yet affect model discovery.

This PR uses that registry to give users a clearer indication of whether Nativ has a candidate loader for a model before they download it.

The classification is deliberately conservative. A model is marked unsupported only when its published configuration provides enough information to reach that conclusion. Models using custom code, ambiguous multimodal configuration, or unfamiliar metadata remain unknown instead of being incorrectly rejected.

## How it works

The Hugging Face API response includes the repository’s configuration metadata, including `model_type`, architecture names, modality configuration, and custom-code indicators.

Nativ compares that metadata with the generated capability registry:

- A matching bundled loader produces a supported result.
- A standard model with an explicit type and task, but no matching loader, produces an unsupported result.
- Missing, custom, or ambiguous configuration produces an unknown result.

The support result affects the badge shown in Discover. It does not prevent downloading or manually attempting to load the model.

For installed models, Nativ keeps both recognized models and unrecognized Hugging Face cache entries visible. More precise separation of complete cached models and interrupted downloads is handled by the next PR.

## Stack

1. Model registry — #409
2. **Model support detection — this PR**
3. Active-download UI

This PR targets `pr/01-model-registry` and depends on #409.

It should remain a draft and must not be merged into the temporary base branch. After #409 merges, this PR will be retargeted to `main` and marked ready for review.

image

### DIFF
--- a/Nativ.xcodeproj/project.pbxproj
+++ b/Nativ.xcodeproj/project.pbxproj
@@ -32,6 +32,7 @@
 		0DF842E2BE07D345918F3D54 /* Manifest.json in Resources */ = {isa = PBXBuildFile; fileRef = 5AF8F6359E4523814FB0EFDA /* Manifest.json */; };
 		0E326825AE5A1A5EEA33D928 /* NativExtensionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = E0DA5C4D0B6B5A094E13CCC8 /* NativExtensionTests.swift */; };
 		0F60E87BC41F29381BD81F73 /* TextDocumentTextExtractors.swift in Sources */ = {isa = PBXBuildFile; fileRef = A91D4332044DCC5DCE4B27F6 /* TextDocumentTextExtractors.swift */; };
+		0F6B875201BCB3195D8E4020 /* HuggingFaceModelSupport.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5734696DFCC890B33C2FC661 /* HuggingFaceModelSupport.swift */; };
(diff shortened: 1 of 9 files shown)


lucasnewman:
@Yavaren Can you add screenshots for the UI changes and tests where applicable?

Also, there are so many conflicts that I need to spend quite a lot of time figuring out how I can integrate it again on top of newly introduced features.

Yavaren:
> @Yavaren Can you add screenshots for the UI changes and tests where applicable?

added tests, the UI adds an orange unsupported badge in discover and makes all Hugging Face cache entries visible in installed models, including previously hidden unrecognized models, which users can attempt to load manually. It improves the UX a lot because users can question why their model is not shown in installed models after downloading it even if it is unsupported. 

We need to figure out what to merge first: this and next PR or my PRs with pinning models (#475 ) and HF link button (#476). Let's merge this one first, and I will resolve all of the conflicts in the smaller RPs I opened today.

PR #411, opened 2026-08-27
show active downloads

Yavaren:
## Summary

This PR shows active model downloads in the Installed Models page while preserving the existing download experience in Discover.

A downloading model appears at the top of Installed Models using the same download-state UI and controls as the Discover page. After the download succeeds, the installed model remains pinned at the top with a temporary green `New` badge.

## What changed

- Added active downloads to the Installed Models page.
- Reused the existing Discover download-state UI and controls.
- Kept the existing top download progress banner unchanged.
- Added a success-only download notification for the UI.
- Added a temporary green `New` badge after successful completion.
- Pinned newly downloaded models above the regular installed-model sorting.
- Removed blue selection outlines from Installed Models rows.
- Prevented the active download and installed model from appearing as duplicate rows.

## Behavior

While a model is downloading:

- The existing progress banner remains visible at the top of the page.
- The model also appears at the top of Installed Models.
- Its row provides the same pause, resume, remove, and progress controls used in Discover.

After a successful download:

- The completed model remains at the top of Installed Models.
- It receives a green `New` badge.
- The temporary badge and pinned position disappear after refreshing or leaving the page.
- Normal installed-model sorting then resumes.

Failed or cancelled downloads are not marked as new.

## Testing

- Built the Nativ macOS application successfully.
- Ran 33 existing model discovery and provider tests with no failures.
- Verified the existing download banner and Discover progress components remain unchanged.
- Verified the rejected cache and retry implementation is absent.

This PR targets `pr/02-model-support` and depends on the preceding PRs in the stack.

Screenshot 2026-08-27 at 7 42 52 PM
Screenshot 2026-08-27 at 7 44 03 PM

### DIFF
--- a/Sources/Nativ/Features/Models/ModelsView.swift
+++ b/Sources/Nativ/Features/Models/ModelsView.swift
@@ -225,6 +225,8 @@ struct ModelsView: View {
     @State private var handledImageModelDiscoveryRequest = 0
     @State private var lastStartedHubSearchTaskID: HubSearchTaskID?
     @State private var readmeSelection: ModelReadmeSelection?
+    @State private var activeDownloadIDs = Set<String>()
+    @State private var recentlyCompletedModelIDs = Set<String>()
 
     init(
         model: NativModel,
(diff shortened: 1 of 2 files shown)


Sources/Nativ/Features/Models/ModelsView.swift
```diff
@@ -490,9 +504,24 @@ struct ModelsView: View {
 
     @ViewBuilder
     private func installedRows(showsResultsHeader: Bool) -> some View {
-        let visibleModels = filteredLocalModels
```
Lazarus-931: i see this filters the installed models, but active downloads bypass the same search, could u as well apply those filters before rendering and counting the active rows too?

lucasnewman (approved):
LGTM

PR #413, opened 2026-08-27
Standardize shared status indicators in Nativ

Lazarus-931:
Standardizes shared status indicators across Nativ to reduce duplicated styling and improve application-wide visual consistency.

This also doesn't change ui significantly, just makes it easier to define ui components and reuse them


lucasnewman (approved):
LGTM

PR #414, opened 2026-08-27
Add file writing, editing, and search tools

lucasnewman:
This adds search (grouped with read permissions) and file writing / editing tools to allow a full workflow for files on disk. Search uses ripgrep (`rg`) under the hood for performance, which I've included in the distribution. The same safeguards are in place as the read side to avoid editing sensitive files without permissions and allowing secrets to enter the context.


Lazarus-931:
hey @lucasnewman this is awesome, is there a demo?

lucasnewman:
Yep, looks like this:

Screenshot 2026-08-27 at 3 58 25 PM

Blaizzy:
This is incredible 🔥🚀

Can't wait!

Blaizzy:
For such cases as in the example (i.e., html) we could later look at rendering it, like in codex for instance

Lazarus-931 (approved):

PR #419, opened 2026-08-28
Add a terminal tool for shell commands

lucasnewman:
All commands currently require explicit approval, and detectable unsafe commands are automatically rejected.

Screenshot 2026-08-28 at 4 05 36 PM

Screenshot 2026-08-28 at 4 10 44 PM


Lazarus-931:
oh i think this still might allow cmds like `/bin/rm -rf /` and `env rm -rf /` to pass won't it?

lucasnewman:
> oh i think this still might allow cmds like `/bin/rm -rf /` and `env rm -rf /` to pass won't it?

There are infinite ways to circumvent the safety guards, which is why we're requiring explicit approval for now. I did expand the pattern matching a bit to try to cover as many cases as is feasible.

Lazarus-931 (approved):

PR #420, opened 2026-08-28
Only include specific document icons for faster build times

lucasnewman:
The PhosphorSwift library adds ~80mb of icons and 75s of additional build time (several minutes in CI). Just including the icons we use reduces the incremental build time to <1s and ~17kb of data.


Lazarus-931 (approved):
this is great, thx!

PR #429, opened 2026-08-30
Reference count system monitor observers

### LINKED ISSUE #349: Support multiple windows

Yavaren:
These stacked PRs close #349 

This pulls the system monitor work out of #372 so it can be reviewed separately. The shared monitor now keeps running as long as at least one view is using it, and closing one view won’t stop updates for another. It also keeps pause and resume behavior consistent across views.


lucasnewman:
@Yavaren Awesome, thanks!

lucasnewman (approved):

PR #430, opened 2026-08-30
Separate shared and window scoped dependencies

Yavaren:
This is the second PR in the multi window stack and builds on #429. Separates application-wide services from state that should belong to an individual window, and adds tests covering which dependencies are shared and which remain independent.


lucasnewman (approved):
👌

lucasnewman (approved):

PR #432, opened 2026-08-30
Synchronize persisted data across windows

Yavaren:
This is the third PR in the multi window stack and builds on #430. It gives each control panel a window ID and uses a shared change hub to keep chat sessions, folders, and image-generation sessions in sync when another window saves or deletes persisted data


lucasnewman (approved):
LGTM

PR #433, opened 2026-08-30
Coordinate inference across windows

Yavaren:
This adds shared inference coordination across control panel windows so the same chat or image generation session cannot be operated on from two windows at once. Sessions that are active elsewhere are temporarily protected from conflicting changes, unrelated sessions can still run independently, and model switching is disabled while any response or image generation is in progress.


lucasnewman:
@Yavaren Could we make InferenceActivityCoordinator @ Observable and use observations for the queue wake-up instead of a PassthroughSubject?

The current way sends synchronously from begin() and can re-enter startNextRequestIfNeeded() before the activeTask is assigned, which could start multiple queued sessions and overwrite the task handle.

Observation would also let it track the coordinator directly, removing the manual objectWillChange calls and duplicated activity state.

lucasnewman (changes requested):
See comment.

Yavaren:
@lucasnewman updated the pr

lucasnewman (approved):
Thanks!

PR #434, opened 2026-08-30
Make artifact deletion transactional

Yavaren:
This makes artifact deletion transactional with its chat or image generation history. An artifact is removed from the cache and index only after its corresponding attachment or generated output is successfully removed and persisted, while locked sessions or persistence failures leave it intact


lucasnewman (approved):
LGTM

PR #435, opened 2026-08-30
Route commands through window intents

Yavaren:
This centralizes navigation from application commands, menu bar actions, extensions, and notifications.


lucasnewman (approved):
Nice!

PR #436, opened 2026-08-30
Enable multiple application windows

Yavaren:
This enables multiple independent Nativ windows by switching the main scene to WindowGroup, adding File > New Window and Shift-Command-N, and giving each window its own navigation and workflow state keeping shared dependencies


lucasnewman:
@Yavaren This looks good but I want to let the other changes bake for one release (should kick off tomorrow) and then we'll light this up for the next release candidate.

lucasnewman (approved):

PR #439, opened 2026-08-30
Fix port reuse when switching models

### LINKED ISSUE #438: Model switching fails because the server port is incorrectly reported as in use
The server stops when switching models but the new model fails to load because the same port is incorrectly reported as still in use, even though no other process is actively using it. The expected behavior is for the new model to start on the same port while still detecting real port conflicts.

Image

Yavaren:
Closes #438 

This fixes the model switching failure where a recently closed server socket could be mistaken for an active port conflict, preventing the next model from loading. Now the port check allows immediate socket reuse, still detecting ports held by active listeners. Regression tests cover both cases.


lucasnewman (approved):
Thanks!

PR #440, opened 2026-08-30
Add external model cache foundation

Yavaren:
This is the first PR in the external model storage stack. It adds the core type used to store and resolve an external model cache location. Selected folders are checked to make sure they are on a local, writable APFS external drive. The reference stores a macOS bookmark so moved folders can be recovered, along with the drive’s volume UUID so Nativ does not use the wrong drive if one appears at the same path.


lucasnewman (approved):
LGTM

PR #446, opened 2026-08-31
Feature/chat import export

Yavaren:
This is the first v of chat import and export.

Exported chats include messages, attachments, model information, the system prompt, and basic metadata. Imported chats get new local IDs and keep previous tool calls as inactive history.

After importing, Nativ shows which model created the chat. Users can switch to that model, download it from the Models page, or continue with another installed model. Chats that exceed the selected model’s context window remain viewable but cannot be continued.

Screenshot 2026-08-31 at 2 50 04 PM
Screenshot 2026-08-31 at 2 44 35 PM
Screenshot 2026-08-31 at 2 44 51 PM

### DIFF
--- a/Docs/features/chat.md
+++ b/Docs/features/chat.md
@@ -27,6 +27,17 @@ carries its title, messages, timestamps, pin state, and an optional folder assig
   session (`folderID`) and the folder list persists alongside sessions.
 - Empty, redundant sessions are pruned automatically.
 
+### Import and export
+
+The action menu for a chat exports a versioned JSON file containing its messages,
+attachments, model repository ID, system prompt, and basic session metadata. Use the import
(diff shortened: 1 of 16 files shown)


Yavaren:
https://github.com/user-attachments/assets/c5d57651-a40e-42d7-a678-873a25b83b18

Lazarus-931:
nice, but i think export exists right?

Yavaren:
@Lazarus-931 yep, it did, but previously it only exported messages in txt. Now it exports:

Chat messages and reasoning
Attachments with their data
Model information
System prompt
Tool-call history
Response metrics
Chat title and timestamps

This way you can basically import the entire conversation and continue it.

lucasnewman (approved):
Nice, looks good!

PR #447, opened 2026-08-31
Add context window progress ring to chat composer

Yavaren:
This adds a context window indicator in chat.

The ring shows how much of the selected model’s context has been used, and hovering over it shows the percentage remaining.

It uses the token count returned by MLX-VLM and the effective context limit already available in Nativ. If runtime metrics aren’t available yet, it falls back to the model’s local metadata.

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatComposer.swift
+++ b/Sources/Nativ/Features/Chat/ChatComposer.swift
@@ -2,6 +2,69 @@ import AppKit
 import Foundation
 import SwiftUI
 
+private struct ChatContextWindowUsage: Equatable {
+    let usedTokens: Int
+    let capacityTokens: Int
+
+    var usedFraction: Double {
+        guard capacityTokens > 0 else { return 0 }
+        return min(max(Double(usedTokens) / Double(capacityTokens), 0), 1)
+    }
+
+    var remainingPercentage: Int {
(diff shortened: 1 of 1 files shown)


Yavaren:
image

Sources/Nativ/Features/Chat/ChatComposer.swift
```diff
@@ -422,6 +489,41 @@ struct ChatComposer: View {
         )
     }
 
+    private var contextWindowUsage: ChatContextWindowUsage? {
+        guard let capacity = effectiveContextWindowCapacity,
+              capacity > 0 else {
+            return nil
+        }
+
+        let usedTokens = viewModel.visibleMessages
```
lucasnewman, line 498: Can you measure how much time this takes to compute for long contexts? This would be run on every View `body` call.

lucasnewman:
@Yavaren Looks cool! See one comment.

Sources/Nativ/Features/Chat/ChatComposer.swift
Yavaren: Yeah it is O(n) but its fine here I think

0.008 milliseconds for 100 messages
0.078 milliseconds for 1,000 messages
and 0.76 milliseconds for 10,000 messages

lucasnewman (approved):

PR #449, opened 2026-08-31
Use lazy stack for chat transcript

Yavaren:
The chat view is basically unusable in long sessions when scrolling/rendering responses. Inference takes about 20 seconds but rendering the same response takes up to 11 minutes with VStack 

Changed from VStask to LazyVStack. 

My biggest PR yet lol :)

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatView.swift
+++ b/Sources/Nativ/Features/Chat/ChatView.swift
@@ -127,7 +127,7 @@ private struct ChatTranscriptView: View {
         let latestUserMessageID = chat.latestUserMessageID
 
         ScrollView {
-            VStack(alignment: .leading, spacing: 12) {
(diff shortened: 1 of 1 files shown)


Yavaren:
Pre-lazy

https://github.com/user-attachments/assets/2f4185ff-02dd-44ea-a05f-70feac351967

Yavaren:
Post-lazy

https://github.com/user-attachments/assets/820332c9-f732-4b54-a22e-f0f0a95f811f

lucasnewman:
This unfortunately breaks the scroll indicator accuracy, but this has come up enough times now that maybe a temporary trade-off is worth it here. It's non-trivial to solve both problems but if we can figure out a way to solve this without dropping back to AppKit that would be a very nice win.

lucasnewman (approved):

PR #455, opened 2026-09-01
Bulk Selection for the Installed Models Page

Lazarus-931:
## Summary

- Add an explicit **Select** action beside Refresh on the Installed Models page.
- Keep bulk-selection controls pinned above the model list so they remain visible while scrolling.
- Support full-row selection, Select All/Deselect All, and confirmed bulk deletion.
- Introduce reusable bulk-selection state and controls for later component-scoped PRs.

Also made search outline slightly darker and removed the divider above it

https://github.com/user-attachments/assets/d005708f-0efb-4747-9c12-7ee3b9ac83ab


lucasnewman (approved):
Looks great, thanks!

PR #462, opened 2026-09-02
Rebuild chat message rendering

Yavaren:
This replaces the previous textual and chunk based chat renderer with a dedicated SwiftUI rendering pipeline built on MarkdownUI. Streaming and completed responses now use the same rendering path, which prevents messages from changing structure when generation finishes.
The renderer supports Markdown, syntax highlighted code, inline and block mathematics, links, tables, lists, and the existing text scaling behavior. Parsed content and highlighted code are cached to avoid repeating expensive work while scrolling or streaming.
Message views retain their natural intrinsic height, allowing very long responses to render completely without clipping.
This also adds tests covering long response height, streaming and completed rendering, and protection against an additional Markdown document background. Existing message actions and chat functionality are unchanged.


project.yml
```diff
@@ -18,6 +18,15 @@ packages:
   textual:
```
lucasnewman, line 18: If we're no longer using Textual we can pull out the dependency.

lucasnewman:
@Yavaren You should merge main since I made some presentation changes to support grouped tool calls but overall this looks like a nice improvement!

Yavaren:
> @Yavaren You should merge main since I made some presentation changes to support grouped tool calls but overall this looks like a nice improvement!

got it, thanks! There will be another PR in this stack in about 30 minutes. If you are running long chats there a bug that makes generation very slow (only ui problem), models that usually run at 250 tokps run at 7tokps in long chats.

The next pr will fix it.

project.yml
Yavaren: But it is still used for model READMEs and audio summaries

lucasnewman: Ok, no worries then.

lucasnewman (approved):

PR #464, opened 2026-09-02
Rebuild chat transcript scrolling

Yavaren:
This rebuilds the transcript around stable message identities, a dedicated bottom scroll target, and explicit tracking of scroll geometry and user interaction. It removes the LazyVStack based transcript and the previous ScrollPosition implementation that could produce incorrect content measurements, clipped messages, and unstable scrolling in long chats.
Automatic scrolling now follows new content only when the user is already near the bottom or when generation is active and the user has not intentionally scrolled away. Manual scrolling pauses automatic movement, returning to the bottom resumes it, and direct navigation to a specific message continues to work.
Streaming scroll updates are isolated in a small observation boundary so each token does not invalidate and rebuild the entire chat page. Message rows remain equatable and use their existing stable identifiers, limiting updates to the message whose content actually changed.
The composer is moved into the normal page layout instead of being positioned over the transcript using continuously measured heights. This removes unnecessary geometry updates and prevents the composer shade from overflowing into generated message parameters.


lucasnewman:
@Yavaren This one needs a merge.

PR #466, opened 2026-09-02
Improved draft model discovery

lucasnewman:
This updates the heuristics for draft model discovery to match `mlx-vlm`, and significantly improves our ability to detect draft models of all types (MTP, DFlash, DSpark, etc)


Lazarus-931 (approved):
thanks this is awesome!

PR #467, opened 2026-09-02
Stack/chat stream relay

Yavaren:
Coalesce high-frequency streaming deltas before main-actor delivery. Publish transcript updates at a smooth 20 Hz cadence. Preserve pending content, reasoning, and the latest generation metrics. Flush pending output on completion and discard it on cancellation.

Users will experience a much better experience during inference in long chats. Previously, small models would generate at 7-10 tok/s in long chats (even though in short chats their performance is about 250 tok/s).


Lazarus-931:
nice, quick question does this changes model performance or the speed of which text renders for very fast models?

Yavaren:
> nice, quick question does this changes model performance or the speed of which text renders for very fast models?

Nope, it changes the render speed. It batches generated tokens and refreshes the chat 20 times per second, so fast models don’t trigger a separate UI render for every token.

lucasnewman (approved):

PR #468, opened 2026-09-02
Add draft model selection to the chat model picker

lucasnewman:
This makes it simple to find and use available draft models for a given model.

https://github.com/user-attachments/assets/424be3fe-d377-42ba-b6b9-d1286be3fe9b


lucasnewman:
If there are no draft models available, a "Find Draft Model..." item appears which jumps to the discover tab filtered to draft models that are published.

PR #469, opened 2026-09-02
Reorder navigation in the sidebar

lucasnewman:
This is based on what I think would be frequency of use for each section, but open to alternative viewpoints.

Screenshot 2026-09-02 at 11 21 00 AM


Lazarus-931:
yeah i think model deserves higher placement as well, and overall the placement looks much better imo

Lazarus-931 (approved):

PR #473, opened 2026-09-02
unsupported model UI badge

Yavaren:
Shows a UI error when an unsupported model is rejected.

image


lucasnewman (approved):

lucasnewman:
@Yavaren Tests are failing

PR #475, opened 2026-09-03
Add model pinning

Yavaren:
This PR adds model pinning to the installed models list.

Pinned models appear in a separate section at the top of the list. Models can be pinned or unpinned using the pin button, context menu, or actions menu. Pins are stored with the existing app settings and removed when a pinned model is deleted.

image


lucasnewman (approved):
Nice!

PR #478, opened 2026-09-03
add HF model support indicators in discovery page

Yavaren:
Replaces the closed PR #410 

This adds an `Unsupported` label to models on the Hugging Face discovery page when Nativ can confidently determine that the bundled MLX runtime cannot load them.

Models with incomplete or ambiguous metadata are left unlabelled rather than being marked unsupported.

Provider detection now also uses `model_type` and `architectures` from the model configuration. The capability manifest generator checks model types against the actual bundled runtime loader.

## Testing

The model support and discovery tests pass. The generated capability manifest was also checked and remains unchanged.

The broader test suite passes apart from an unrelated shell-script timeout. That test passes when run separately.

The UI of the model discover page shows this:

image


Lazarus-931:
would this work with #473?

Yavaren:
@Lazarus-931 yes, this PR only adds the unsupported badge for the model in the discovery page, does not touch models in installed

lucasnewman:
@Yavaren I took a quick look, can you see if we can fix the issues pointed out by these tests before we land this? `model_type` should be authoritative at the very least.

```swift
    func testUnsupportedPrimaryModelTypeTakesRuntimePrecedence() throws {
        let result = try makeClassifier().classify(
            configuration: configuration([
                "model_type": "unrecognized-wrapper",
                "speculators_model_type": "drafter-loader",
            ]),
            pipelineTag: "text-generation",
            tags: []
        )

        XCTAssertEqual(result, .unsupported)
    }

    func testMalformedArchitecturesRemainUnknown() throws {
        let result = try makeClassifier().classify(
            configuration: configuration([
                "model_type": "not-in-the-bundled-runtime",
                "architectures": "BoundaryExtractor",
            ]),
            pipelineTag: "text-generation",
            tags: []
        )

        XCTAssertEqual(result, .unknown)
    }
```

lucasnewman (changes requested):
See comment about cases that need to be fixed.

Yavaren:
> See comment about cases that need to be fixed.

please have a look at the latest commit

Yavaren:
@lucasnewman resolved merge conflicts

lucasnewman (approved):

lucasnewman:
Looks good!

PR #480, opened 2026-09-03
Keep bulk selection scoped to visible items

Lazarus-931:
Keeps Models and Artifacts bulk actions scoped to visible filtered items while moving Artifacts onto the shared selection state and toolbar.

For consistency, why have two separate select states

### DIFF
--- a/Sources/Nativ/Features/Models/ModelsView.swift
+++ b/Sources/Nativ/Features/Models/ModelsView.swift
@@ -356,6 +356,9 @@ struct ModelsView: View {
         .onChange(of: localLibrary.models.map(\.repoID)) { _, _ in
             selectFirstVisibleReadmeIfNeeded(in: .installed)
         }
+        .onChange(of: eligibleVisibleModelIDs) { _, visibleIDs in
+            installedModelSelection.retain(visibleIDs)
+        }
         .onChange(of: hubLibrary.models.map(\.id)) { _, _ in
             selectFirstVisibleReadmeIfNeeded(in: .discover)
         }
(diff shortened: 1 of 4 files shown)


Sources/Nativ/Features/Models/ModelsView.swift
```diff
@@ -356,6 +356,9 @@ struct ModelsView: View {
         .onChange(of: localLibrary.models.map(\.repoID)) { _, _ in
             selectFirstVisibleReadmeIfNeeded(in: .installed)
         }
+        .onChange(of: eligibleVisibleModelIDs) { _, visibleIDs in
+            installedModelSelection.retain(visibleIDs)
+        }
```
Yavaren: /Users/yavaren/Documents/nativ/Sources/Nativ/Features/Models/ModelsView.swift:278:25 The compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions

Swift exceeded the type-checking complexity limit for ModelsView.body reported at line 278

Yavaren (commented):
Check my comment

Sources/Nativ/Features/Models/ModelsView.swift
Lazarus-931: hm yeah will move it to smaller Installed section

lucasnewman (approved):

Sources/Nativ/Features/Models/ModelsView.swift
Yavaren: Works now, LGTM

PR #488, opened 2026-09-04
Fix latest prompt editing without branching

### LINKED ISSUE #487: Prompt edits create indistinguishable duplicate-looking chats
## Summary

Editing and resending the latest user prompt creates a second persisted chat session, but the original and the branch are indistinguishable in the sidebar. Users perceive this as Nativ duplicating chats even when they never explicitly select **Fork chat from this response**.

## Reproduction

1. Start a chat and complete at least one response.
2. Use the edit action on the latest user prompt.
3. Change or resend the prompt.
4. Return to the sidebar.

## Actual behavior

A new sidebar row appears with the same title as the original chat. The original is retained, and neither row is marked or grouped as a branch.

## Cause

In `ChatViewModel.send()`, a non-nil `promptEditContext` calls `ChatPromptRevision.make`, then unconditionally creates a new `ChatSession` through `ChatConversationBranch.make` and persists it through `activateBranch`.

The branch:

- receives a new session UUID;
- copies the original transcript prefix and message IDs;
- sets `customTitle` to `nil`;
- derives its title from the same first user message;
- stores no parent-session/branch provenance for the sidebar.

Consequently, multiple distinct session files render as identically named chats.

This is not a `ForEach` identity problem or duplicated loading: persisted session IDs are unique.

## Expected behavior

Prompt editing should not create an unexplained duplicate-looking chat. Either:

1. edit/regenerate within the current chat, or
2. retain branching but persist parent/branch metadata and clearly group or label branches in the sidebar.

At minimum, the edit UI and resulting sidebar row should make the branch relationship unmistakable.

Tracking: [NEY-147](https://linear.app/neywalabs/issue/NEY-147/prompt-edits-create-indistinguishable-duplicate-looking-chats)

Yavaren:
Closes #487 

This pr fixes editing so that editing the latest user message updates the current chat and regenerates the model’s response without creating a new branch

### DIFF
--- a/Sources/Nativ/Features/Chat/ChatComposer.swift
+++ b/Sources/Nativ/Features/Chat/ChatComposer.swift
@@ -1082,7 +1082,7 @@ struct ChatComposer: View {
             return blockingNotice.message
         }
         if viewModel.promptEditContext != nil {
-            return "Fork and regenerate (Return)"
+            return "Save and regenerate (Return)"
         }
         return "Send (Return)"
     }
@@ -1131,7 +1131,7 @@ private struct ChatPromptEditBanner: View {
             VStack(alignment: .leading, spacing: 2) {
                 Text("Editing prompt")
                     .fontWeight(.medium)
-                Text("Sending will create a new chat branch.")
+                Text("Sending will replace the latest response.")
                     .foregroundStyle(.secondary)
             }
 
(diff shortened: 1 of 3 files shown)


Lazarus-931:
tested it out, thanks!

PR #490, opened 2026-09-04
Add progressive tool exposure and discovery

Lazarus-931:
## Summary

- add Off, Auto, and On exposure modes for built-in, custom, and MCP tools
- replace repeated segmented selectors with one colored access button per row: gray Off, blue Auto, and green On
- keep essential direct tools On while migrating other enabled tools and MCP servers to Auto
- include enabled file and terminal tools directly in Projects, preserve Off, and let users re-enable disabled essentials from Projects settings
- expose Tool Search as a Built-in tool that defaults to On and can itself be set to Off, Auto, or On
- make the single tool_search capability search Nativ existing built-in definitions, saved custom tools, and live MCP tools without introducing a second tool registry
- expose matching Auto schemas only for the current request and enforce exposure again at execution time

## Verification

- xcodegen generate
- latest focused project/tool/settings regression checks: 174 passed, 0 failed
- full-suite follow-up: one unchanged Buzz setup test fails three assertions; three opt-in real-model tests skipped
- final unsigned Debug build with Xcode 26.6
- full Nativ test suite on the initial implementation
- live LFM2.5-8B-A1B test: initial standalone request dropped from 8,223 to 1,033 total tokens
- live discovery loop: tool_search found git_status, exposed it on the next round, and dispatched it
- visually checked the compact controls, Auto magnifying-glass state, hover/accessibility values, unconfigured Set Up rows, and Terminal default On state

Currently, the tools page is tweaked to show this instead, where there is a Off/Auto/On(Gray/Blue/Green) switch 

Screenshot 2026-09-04 at 12 09 42 PM
ble.


Lazarus-931:
Significant drop in context model is exposed to, while still allowing it to use the tool-search tool built in(auto-on), to find 'discoverable' tools:

Below, same question resulted almost 8x less context model is exposed to, still testing WIP

Screenshot 2026-09-04 at 12 07 53 PM

Lazarus-931:
Further testing shows, models still have the ability to discover tools, while overall context they are exposed to per user prompt decreases massively.

Screenshot 2026-09-05 at 10 43 11 AM

Screenshot 2026-09-05 at 10 43 29 AM


drops to a 1k context from 8k(https://github.com/Blaizzy/nativ/pull/490#issuecomment-5543260892)

lucasnewman:
@Lazarus-931 Very cool! It's still marked as a draft, is this ready for review?

lucasnewman:
@Lazarus-931 I think we should still have Projects auto-include the file search/read/write/patch and terminal tools, since that's kind of the point in that context. With this change auto mode would force a discovery call first and off would actually remove the tool usage which defeats the feature.

lucasnewman:
Also looks like one concurrency issue:

```sh
nativ/Sources/Nativ/Features/Extensions/MCPSectionView.swift:225:30: warning: converting non-Sendable function value to '@isolated(any) @Sendable (ToolExposureMode) -> Void' may introduce data races
                        set: onExposureModeChange
                             ^

nativ/Sources/Nativ/Features/Extensions/MCPSectionView.swift:225:30: Converting non-Sendable function value to '@isolated(any) @Sendable (ToolExposureMode) -> Void' may introduce data races
```

lucasnewman:
In a new project, terminal tool discovery is failing:

Screenshot 2026-09-05 at 10 42 30 AM

Blaizzy:
This is cool!

I'm just confused with these icons.

Does it mean search? Or view the tool? 


It's if auto/discovery, we should look for a better icon

 image

Lazarus-931:
> This is cool!
> 
> I'm just confused with these icons.
> 
> Does it mean search? Or view the tool? 
> 
> 
> It's if auto/discovery, we should look for a better icon
> 
>  image

yes, means it's a tool discoverable by the tool search, will update

PR #494, opened 2026-09-04
Unify Markdown rendering for model cards

lucasnewman:
We switched chat over to a separate implementation -- this removes Textual and unifies the implementation for the model cards as well.

Before:
Screenshot 2026-09-04 at 8 28 22 AM

After:
Screenshot 2026-09-04 at 8 34 19 AM


Lazarus-931 (approved):
nice!

PR #496, opened 2026-09-04
Fix model configuration header overlap

Lazarus-931:
Reserve space in the model configuration header so reset and sidebar controls remain visually distinct.

top right here: 
Screenshot 2026-09-04 at 12 49 44 PM


Blaizzy (approved):
LGTM!

PR #500, opened 2026-09-05
Fix model picker tooltip clipping

Lazarus-931:
- present the composer model picker tooltip through the shared arrowless popover
- keep long disabled-state guidance visible outside the composer's clipped bounds
- reuse the same behavior for chat and image model pickers

before:
image

after:

image


Lazarus-931:
https://github.com/user-attachments/assets/ee8cceb1-48e1-4b6e-be22-900b7d813e33

lucasnewman (approved):

PR #507, opened 2026-09-05
Clean up sidebar layout and interactions for Projects and Sessions

lucasnewman:
Fix up alignment, layout, icons, and animations in the sidebar, and simplify the interactions for creating the chat and editing the lists to be consistent across the board.

This also prevents new chats from having the ugly timestamp name initially and fixes and issue where the field editor to renaming a chat could persist even when selecting or creating a different chat.

https://github.com/user-attachments/assets/8513e51e-bd22-48c3-a2c8-cef85e3b7857


Lazarus-931 (approved):
lgtm

PR #512, opened 2026-09-06
Refine sidebar navigation, pinned chats, and bulk selection

Yavaren:
Adds New chat and Import chat actions, and places bulk controls inside Sessions. Aligns folder/project hover buttons. 

image
image


lucasnewman:
Why would "Import Chat" need this much prominence in the UI?

lucasnewman:
Also we shouldn't drop the compose button next to the sections header, that just makes it harder to use. The New Chat and other changes seem good.

Yavaren:
> Also we shouldn't drop the compose button next to the sections header, that just makes it harder to use. The New Chat and other changes seem good.

I recreated the codex behaviour in the side nav per prince's feedback. It seems to be a more popular placement for users.
image

increased contrast too
image

lucasnewman (approved):

PR #520, opened 2026-09-07
Safer voice activation key combo monitoring

### LINKED ISSUE #511: EXC_BAD_ACCESS in FnControlShortcutMonitor MainActor executor check
## Summary

Nativ 0.3.6 intermittently crashes on the main thread while the built-in Audio extension polls the global voice shortcut modifier keys.

The crashing path is the repeating timer in `FnControlShortcutMonitor.startModifierPolling()`. Its callback invokes `MainActor.assumeIsolated`, and the process dies while the Swift runtime validates the main-actor executor:

~~~text
objc_opt_class + 48
swift_getObjectType + 204
swift_task_isMainExecutorImpl + 36
swift::SerialExecutorRef::isMainExecutor() + 24
swift_task_isCurrentExecutorWithFlagsImpl + 72
Nativ + 3619328
Nativ + 3628124
Nativ + 1656052
__NSFireTimer + 104
~~~

Exception:

~~~text
EXC_BAD_ACCESS (SIGSEGV)
KERN_INVALID_ADDRESS at 0x000000000000001e
Termination Reason: SIGNAL, Code 11, Segmentation fault
Triggered by Thread 0: com.apple.main-thread
~~~

This is a native arm64 process (`translated: false`). The termination is not an out-of-memory kill and does not originate in the model server.

## Environment

- Nativ: 0.3.6 (build 202608311942)
- macOS: 26.5.2 (25F84)
- Hardware: Mac Studio (Mac15,14)
- Chip: Apple M3 Ultra
- Unified memory: 512 GB
- Audio / Voice Dictation extension: enabled
- Architecture: arm64 native

## Steps to reproduce in Nativ

1. Install Nativ 0.3.6.
2. Enable the built-in **Audio** extension, leaving the voice modifier shortcut enabled.
3. Quit Nativ completely and relaunch it.
4. Leave the app running, optionally exercising the Fn+Control shortcut.
5. The 60 Hz modifier polling timer eventually fires into the executor-check path above and the app terminates.

The failure appears intermittent, so the exact delay may vary. In the captured occurrence, the app crashed roughly 33 seconds after launch without evidence of a model request causing it.

## Minimal reduction

The following standalone macOS app repeatedly exercises the same Timer-to-`MainActor.assumeIsolated` transition. Build it with the same Swift/macOS toolchain as Nativ:

~~~swift
import AppKit
import ApplicationServices
import Foundation

@main
@MainActor
final class AppDelegate: NSObject, NSApplicationDelegate {
    private var modifierPollTimer: Timer?

    static func main() {
        let app = NSApplication.shared
        let delegate = AppDelegate()
        app.delegate = delegate
        app.run()
        withExtendedLifetime(delegate) {}
    }

    func applicationDidFinishLaunching(_ notification: Notification) {
        let timer = Timer(timeInterval: 1.0 / 60.0, repeats: true) { _ in
            MainActor.assumeIsolated {
                _ = CGEventSource.flagsState(.combinedSessionState)
            }
        }
        RunLoop.main.add(timer, forMode: .common)
        modifierPollTimer = timer
    }
}
~~~

On systems where the standalone reduction does not fail, enabling Audio and relaunching Nativ is the application-level reproduction; another component may be corrupting the executor reference before this check exposes it.

## Source mapping

The unsymbolicated Nativ frame was mapped by disassembling the v0.3.6 binary. The function contains the literal source path `Nativ/FnControlShortcutMonitor.swift` and calls the shared `MainActor.assumeIsolated` helper. It matches:

https://github.com/Blaizzy/nativ/blob/v0.3.6/Sources/Nativ/Features/VoiceCapture/FnControlShortcutMonitor.swift#L252-L261

## Expected behavior

The Audio shortcut monitor should run continuously without terminating the app.

lucasnewman:
Use an async task for monitoring instead of a timer and avoid use of `assumeIsolated()` which is unsafe in this context. Resolves https://github.com/Blaizzy/nativ/issues/511

### DIFF
--- a/Sources/Nativ/Features/VoiceCapture/AudioCaptureLibrary.swift
+++ b/Sources/Nativ/Features/VoiceCapture/AudioCaptureLibrary.swift
@@ -115,7 +115,7 @@ final class AudioCaptureLibrary: ObservableObject {
     private let meetingJoinMonitor = MeetingJoinMonitor()
     private let meetingSuggestion = MeetingTranscriptionSuggestionController()
     private let analytics: AudioAnalyticsStore
-    private var elapsedTimer: Timer?
+    private var elapsedTask: Task<Void, Never>?
     private var captureStartedAt: Date?
     private var shouldSummarizeCurrentCapture = false
     private var activeBackend: ActiveAudioCaptureBackend?
@@ -270,7 +270,7 @@ final class AudioCaptureLibrary: ObservableObject {
             phase = .recording
             recordingOverlay.update(level: 0, elapsed: 0)
             recordingOverlay.didStartRecording()
-            startElapsedTimer()
+            startElapsedUpdates()
         } catch {
             if Self.isScreenCapturePermissionError(error) {
                 // ScreenCaptureKit presents the native permission dialog itself.
@@ -305,7 +305,7 @@ final class AudioCaptureLibrary: ObservableObject {
 
         phase = .processing
         recordingOverlay.waitForTranscription()
-        stopElapsedTimer()
+        stopElapsedUpdates()
         let duration = max(
             elapsed,
             captureStartedAt.map { Date().timeIntervalSince($0) } ?? 0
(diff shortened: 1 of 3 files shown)


Blaizzy (approved):
LGTM, thanks!

PR #524, opened 2026-09-08
Recall the last prompt with the Up arrow

Lazarus-931:
Pressing Up in an empty composer loads your last prompt back in for editing, reaching the edit flow that already existed without hunting for the small button on the sent message.

- Recall only happens from an empty composer, so Up never stops being Up while there is text to move through and the gesture cannot discard something half-written; any modifier passes through, and an Up that recalls nothing still moves the caret.
- A `↑ to show last` hint appears in the placeholder only while the gesture would work, and retires permanently the first time it is used.
- The editing banner is calmer and shorter: no accent colour, standard separator border, and one line saying what sending does rather than two.

https://github.com/user-attachments/assets/7bc770d0-3ed5-40a1-8c9a-d015113e0f05


Sources/Nativ/Features/Chat/ChatComposer.swift
```diff
@@ -1177,49 +1198,60 @@ struct ChatComposer: View {
...
             Image(systemName: "pencil")
-                .foregroundStyle(Color.accentColor)
 
-            VStack(alignment: .leading, spacing: 2) {
-                Text("Editing prompt")
-                    .fontWeight(.medium)
-                Text("Sending will replace the latest response.")
-                    .foregroundStyle(.secondary)
-            }
+            Text("Replace the last response")
```
lucasnewman: Should this be "Replace the last message"? I tend to think of "response" as the model response, not what I wrote.

Lazarus-931: yes, that makes sense, will change :)

Nativ.xcodeproj/xcshareddata/xcschemes/Nativ.xcscheme
```diff
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "2660"
-   version = "1.3">
+   LastUpgradeVersion = "1430"
```
lucasnewman: Ideally we don't go backwards here, or Xcode will put a bunch of "Recommended Project Changes" things into the Warnings & Errors panel, some of which aren't helpful/correct. Can we revert this?

I know `xcodegen` is the one that overwrites it -- maybe we should also add it to `.gitignore` so it's just a local reference.

lucasnewman (approved):
Looks good, see comments though

PR #525, opened 2026-09-08
Add quote replies to chat messages

Yavaren:
You can select text in a user or assistant message and click Quote reply to attach it to your next message. The selected passage appears in the composer, where you can remove it or jump back to the original message.

The selected text is sent along with your message. Quotes are preserved when editing, branching, and importing or exporting chats. This also fixes window dragging getting in the way of text selection.

Selection currently stays within one Markdown block, such as a paragraph or heading.

image
image
image


Sources/Nativ/Features/Chat/ChatView.swift
```diff
@@ -262,6 +262,9 @@ private struct ChatTranscriptView: View {
                 latestUserMessageID: latestUserMessageID,
                 forkableAssistantResponseIDs: forkableAssistantResponseIDs
             )
+            .environment(\.chatAnnotationAction, chat.addAnnotation)
+            .environment(\.navigateToChatAnnotation, { chat.scrollTargetMessageID = $0 })
```
lucasnewman: Setting closures as part of view construction is an anti-pattern in SwiftUI, because there's no way to have proper equality checks work in this case and we'll get excessive re-rendering of the view bodies.

Yavaren: Hm thanks, will fix this.

Yavaren: How is everything else overall?

lucasnewman: The feature looks cool! If we can make sure we don't induce any performance regressions I'll take a deeper look.

Yavaren: @lucasnewman the fix is out

lucasnewman:
@Yavaren I introduced some conflicts with the Markdown changes. I went ahead and merged main into your branch and updated the implementation to make sure it handled escaped characters (like &amp; -> &) correctly.

Would you mind testing it out and making sure it looks ok? I'll do some testing on my side as well.

Yavaren:
> @Yavaren I introduced some conflicts with the Markdown changes. I went ahead and merged main into your branch and updated the implementation to make sure it handled escaped characters (like & -> &) correctly.
> 
> Would you mind testing it out and making sure it looks ok? I'll do some testing on my side as well.

Overall, really great improvement in terms of scrolling experience, the only thing I would want to fix is allowing users to select multiple sections of markdown at once. Right now, if you want to quote reply, you can only select a single piece of markdown

https://github.com/user-attachments/assets/31edf4cb-290a-4077-b4ed-585d02349b17

Because users would definitely want to select multiple sections of text.

Yavaren:
@lucasnewman there is another improvement we could make to the scrolling experience. 


https://github.com/user-attachments/assets/753d885c-0348-4729-8169-2c3a73d472d7

Do you think we can make this reliable?

lucasnewman:
@Yavaren I pushed some fixes for math content and a performance change to make sure we don't spend cycles rendering a massive selection block outside of the visible area. In my testing it looks good now!

lucasnewman (approved):

Yavaren:
@lucasnewman thanks, next time will pay attention to this

PR #527, opened 2026-09-08
Stop tracking the generated Xcode scheme

Lazarus-931:
per rev in #524, to prevent xcode from surfacing proj changes that aren't correct, this pr gitignores the keeps as a local reference, which nothing depends on since both CI workflows run xcodegen before xcodebuild.


lucasnewman (approved):
Thanks!

PR #533, opened 2026-09-10
Resolve exact model download sizes

Yavaren:
Discovery previously estimated download sizes from tensor metadata, with a fixed multiplier for image models. This omitted supporting files and alternative weight formats selected by the downloader. For example, `sentence-transformers/all-MiniLM-L6-v2` appeared as approximately 90.9 MB despite its selected files totaling 976.9 MB.

This change calculates sizes from the exact selected-file metadata, excluding GGUF files according to the existing download policy **(Prince will GGUF back later).** 

**Changes**

- Display verified download totals and use them for size sorting.
- Resolve sizes asynchronously with a 250 ms debounce, up to four concurrent requests, shared requests, cancellation, and caching.
- Show “Checking size…” or “Size unavailable” instead of an inaccurate estimate.
- Pin downloads started from Discover to the displayed repository revision.
- Verify the download file list and remaining disk space before transferring files, accounting for cached files and other download reservations.
- Keep runtime memory estimates separate from download sizes.

**Model verification**

All 10 models below displayed the expected totals in the updated discovery UI. Previous estimates were calculated from the original implementation; corrected totals were verified against recorded repository file metadata.

| Model | Previous estimate | Verified total |
| --- | ---: | ---: |
| `mlx-community/whisper-tiny-asr-fp16` | 74.4 MB | 78.8 MB |
| `sentence-transformers/all-MiniLM-L6-v2` | 90.9 MB | 976.9 MB |
| `sentence-transformers/all-MiniLM-L12-v2` | 133.4 MB | 1.30 GB |
| `BAAI/bge-small-en-v1.5` | 133.4 MB | 401.1 MB |
| `openai/whisper-tiny` | 151.0 MB | 608.9 MB |
| `intfloat/e5-small-v2` | 133.4 MB | 803.2 MB |
| `BAAI/bge-base-en-v1.5` | 437.9 MB | 1.31 GB |
| `sentence-transformers/all-mpnet-base-v2` | 437.9 MB | 3.82 GB |
| `stabilityai/sdxl-turbo` | 25.67 GB | 55.52 GB |
| `stabilityai/stable-diffusion-xl-base-1.0` | 25.67 GB | 76.91 GB |

These are full selected-file totals before subtracting cached files, including alternative weight formats selected by the current downloader. 

**Performance benchmarks**

Both builds completed the same 45-operation workload on the same Mac, running under Xcode in Debug with a 1920 × 1080 window. Search timings are medians of five trials; entry and pagination timings use three trials.

| Operation | Before | After |
| --- | ---: | ---: |
| Open Discover | 240 ms | 264 ms |
| Search: empty query | 632 ms | 790 ms |
| Search: `mlx-community` | 520 ms | 598 ms |
| Search: `sentence-transformers` | 510 ms | 542 ms |
| Search: `BAAI/bge` | 535 ms | 651 ms |
| Search: `stabilityai/sdxl-turbo` | 560 ms | 696 ms |
| Page 1 → 2 | 551 ms | 557 ms |
| Page 2 → 3 | 327 ms | 343 ms |
| Page 3 → 2 | 298 ms | 318 ms |
| Page 2 → 1 | 304 ms | 324 ms |
| Mean scrolling CPU | 13.47% | 16.38% |
| Final idle CPU | 2.41% | 2.37% |
| Peak sampled memory footprint | 561.7 MB | 528.8 MB |

These timings measure when search results appear, not when every size lookup completes. Results come from warm sessions against the live Hub.

### DIFF
--- a/Docs/features/models.md
+++ b/Docs/features/models.md
@@ -68,6 +68,13 @@ transfer, Nativ verifies the file list and checks the additional space needed af
 cached files and reservations for other downloads. If that verification fails, the download
 reports an error instead of proceeding with an unknown total.
 
+After resolving its exact uncached size, each downloader waits for the app to approve a disk
+reservation. A shared lock makes the fresh capacity check and reservation atomic across
+downloads on the same filesystem. Reservations remain held while paused and until the
+subprocess exits, including on cancellation or failure. Retries resolve and reserve again.
+The full reservation is retained during each attempt because displayed progress can be
+interpolated; this conservatively leaves less space available for additional downloads.
+
 ## Roles and preloading
 
 Separate models can be assigned per role and loaded concurrently, memory permitting:
--- a/Docs/features/models.md
+++ b/Docs/features/models.md
@@ -13,6 +13,12 @@ Installed models are found by scanning the Hugging Face cache plus any additiona
 - Capabilities are derived per model from its `config.json` (or `model_index.json` for
   diffusion pipelines) — see [`LocalModelDiscovery`](../../Sources/Nativ/Features/Models/LocalModelDiscovery.swift).
 
+Discover shows the total size of the files selected for download, including tokenizers and
+additional weight formats in the repository. Optional GGUF files are excluded. Sizes load in
+the background for visible rows; “Checking size…” changes to a verified total or “Size unavailable.”
+Size sorting uses these totals across the buffered search results. Memory-fit warnings remain
+separate estimates of runtime memory needs.
+
 ## External model storage
 
 The primary Hugging Face cache can live on a locally attached APFS external drive. Open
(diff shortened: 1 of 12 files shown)


Lazarus-931:
awesome work @Yavaren, i guess accuracy is more important than the ui latency here, but I wonder if the ms regressions can be recouped, was this difference noticeable when you tested?

Yavaren:
@Lazarus-931 thanks a lot! I don't see a noticeable difference. I had a latency concern when I started building the feature, so I included benchmarks in case you had it as well.

lucasnewman:
@Yavaren The one issue I see is that disk size reservation can race with multiple requests before the sizes are resolved. It would be better to have the size estimation resolve first, then reserve capacity in some atomic way (a lock or whatnot).

Yavaren:
> @Yavaren The one issue I see is that disk size reservation can race with multiple requests before the sizes are resolved. It would be better to have the size estimation resolve first, then reserve capacity in some atomic way (a lock or whatnot).

Got it, will add soon.

Yavaren:
@lucasnewman any other issues?

Yavaren:
> @Yavaren The one issue I see is that disk size reservation can race with multiple requests before the sizes are resolved. It would be better to have the size estimation resolve first, then reserve capacity in some atomic way (a lock or whatnot).

fixed

lucasnewman (approved):
Looks good, thanks!

Lazarus-931 (approved):

PR #541, opened 2026-09-14
Add pasted text attachments to chat

Yavaren:
This PR makes pasted text over 2,000 characters long an attachment in the UI. It does not make the text an actual attachment, and from the API standpoint, it is regular pasted text. However, if we follow this path of displaying long texts as attachments, we have way less Markdown to render, which will improve the scrolling experience. After pasting text, yon can open it in a popup and read and copy the contents. 

image
image


Lazarus-931 (approved):

lucasnewman:
@Yavaren This needs follow-ups for correctness issues:

P1 — Undo can crash after IME input. Committing 你 through marked text, then undoing twice, crashes with NSRangeException. The draft snapshot and native text undo occupy separate groups, leaving a native undo operation targeting text already removed. The pre-merge editor passes the same sequence.

P2 — Typing duplicates large attachments in undo history. Every edit rebuilds the full message and retains its previous value. A 1 MB paste followed by 100 characters added approximately 100 MB in an optimized fixture. Keeping attachment content separate and assembling the message when sending would simplify this and avoid the duplication.

Yavaren:
@lucasnewman thanks, will patch asap.

PR #546, opened 2026-09-14
Add configurable spoken Return command for dictation

Blaizzy:
Dictation can now press Return when speech ends with a configured word or phrase. For example, “Send me the details enter” inserts “Send me the details” and then presses Return, omitting the trigger from the clipboard, saved transcript, and history. The behavior applies to both speech engines and retries; a command-only dictation presses Return without pasting or changing the clipboard.

Audio → Shortcuts now includes a Spoken Return toggle, an editable trigger word or phrase, and Restore Default. The command defaults to enabled with “enter”; disabling it or leaving the trigger blank preserves ordinary dictated text. Preferences persist across launches and preserve existing shortcut settings. Matching handles capitalization and trailing punctuation, treats custom triggers literally, and only recognizes the final phrase.

### Shortcuts page

Audio Shortcuts page with the Spoken Return toggle and editable trigger

Enable or disable **Spoken Return**, choose a trigger word or phrase, or restore **“enter”**. The trigger is removed only when it ends a dictation; meeting recordings and their transcripts are unaffected.

Validation:
- macOS Debug build succeeded; signed app launched and the Shortcuts page was visually checked.
- All 15 focused parser and preference tests passed in an isolated SwiftPM harness.
- The Xcode test target remains blocked by the existing `SystemMonitorIdentityTests.swift` import of the unavailable `Nativ` module.
- Live microphone-to-Return behavior has not been manually exercised.


lucasnewman (approved):
LGTM

PR #548, opened 2026-09-14
Fix composer undo crash and attachment memory growth

Yavaren:
Addresses both follow-up issues Lucas reported in [PR #541](https://github.com/Blaizzy/nativ/pull/541#issuecomment-5667227104).

- Fixes the IME undo crash: committing marked text such as `你` and undoing twice no longer leaves a native undo operation targeting text already removed by draft undo. Draft history owns undo for the attachment-aware composer, with Undo/Redo commands retained.
- Fixes attachment duplication in undo history: editable text and attachment content are stored separately, and the complete message is assembled when sending. Draft snapshots share attachment content while preserving attachment positions, original preview text, and send-time trimming.

Validation: the app builds and all 21 targeted Xcode tests pass. Four additional editor regression checks pass in an isolated AppKit/SwiftUI harness, including the original IME crash sequence. Manual checks covered a 1 MiB paste, 100 typed characters, IME input, and repeated undo/redo without crashes.

In the optimized 1 MiB attachment + 100 edits fixture, retained memory growth fell from approximately 103 MiB to 0.19 MiB. All 100 undo and 100 redo operations preserved the attachment content.


lucasnewman (approved):
Thanks!

PR #550, opened 2026-09-14
Avoid update errors during publish

lucasnewman:
Temporarily publish the existing appcast.xml during the publish action for releases, so the app doesn't show transient update errors.


Lazarus-931 (approved):

Blaizzy (approved):
LGTM

PR #551, opened 2026-09-14
Personalized memories

Yavaren:
Add chat personalization and optional profile export

Adds a Personalization page in Settings where users can enter their preferred name, occupation, and information about themselves.

Users can choose a chat style: Default, Concise, Friendly, Professional, or Detailed. They can also control emoji usage with None, Default, or More, and choose Minimal, Default, or Structured Markdown formatting.

Personalization applies to new chats. Each chat keeps the personalization it started with, so changing these settings does not change existing conversations.

Chat exports exclude personalization by default. An “Include personalization” checkbox in the save dialog lets users include their saved profile and response preferences. The checkbox starts unchecked for every export. If included, personalization is restored when the chat is imported.

When exporting, users have an option to include/exclude their personalization 
image

Validated with 98 passing tests and checks of the save dialog, including both checkbox states and the default reset behavior.

image
image


Sources/Nativ/NativSettings.swift
```diff
@@ -405,6 +405,214 @@ struct ModelConfigProfile: Codable, Equatable {
...
+            }
+        }
+
+        var systemPrompt: String {
+            let instructions: String
+            switch self {
+            case .default:
+                return ""
+            case .concise:
+                instructions = """
```
Lazarus-931, line 436: hm, I think this is a fair pr, but i hope this doesn't crush small models, can u test to verify that given lengthy responses + this hard coded prompt, model doesn't degrade or loose focus. 

ideally I don't for see a problem with this, but just worth mentioning :)

lucasnewman:
Overall this looks good. I think we can go ahead and remove the "Default —" prefix from these, it doesn't add anything.

Screenshot 2026-09-16 at 8 51 26 AM

Yavaren:
@lucasnewman I removed the prefixes completely, I agree, they don't make much sense 
Screenshot 2026-09-18 at 12 06 45 PM
Screenshot 2026-09-18 at 12 06 52 PM

lucasnewman (approved):
Looks good!

Lazarus-931 (approved):

Sources/Nativ/NativSettings.swift
Yavaren: Sure, I will wrap something up and will test things to show you

Yavaren: Smaller models don't really follow instructions. I tested on 0.5 and 0.8 B qwen models. Also tried gemma-3-270m and the same problem. The model starts telling me about itself when I ask it who am I. 

The only exposed issue is the emoji prompt, it starts working only when you ask the model in the chat. I will update the prompt and merge the PR, it is ready.

Issue #547, opened 2026-09-14
Getting a model fit warning for models that will actually fit


roehnan:
### Category
App UI issue

### What happened
Model fit during download not using the actual size: 

<img width="565" height="193" alt="Image" src="https://github.com/user-attachments/assets/9793c4ca-54a6-4e36-b13e-3ed80d306f11" />

<details>
<summary>Diagnostics</summary>

**Environment**
- Nativ: 0.3.8rc1 (202609111702)
- macOS: macOS 26.6.2 (Build 25G83)
- Chip: Apple M3 Max
- Memory: 96 GB total, 25.31 GB in use
- mlx-vlm: 0.7.0

**Model & Server**
- Server: running
- Selected model: On demand
- Loaded model: None
- Max output tokens: 2048
- Context window: model default
- KV quantization: off
- Speculative decoding: off
- Prefix caching: off
- Thinking: on
- Launch arguments: --host 127.0.0.1 --port 8000 --max-tokens 2048
- Pending settings change: server restart required

</details>

<details>
<summary>Server output (last 5 lines)</summary>

```
Started mlx-vlm-server.
INFO:     Started server process [88758]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL C to quit)
```

</details>

Lazarus-931:
hi @roehnan, https://github.com/Blaizzy/nativ/pull/573 was merged into main

should address the issue, closing!

thx for the bug report :)

PR #554, opened 2026-09-14
Account for quantized weight packing in model size estimates

Lazarus-931:
Solved #547, was something I did before that messed up size calculation.

This change makes quantized packing(how model releases store weights in hf), something `HuggingFaceHub.swift` could understand to prevent it.

Same device(128gb)

## Before(What issue saw):

Screenshot 2026-09-14 at 7 42 08 PM

## After(this pr):
Screenshot 2026-09-14 at 7 42 30 PM


Yavaren:
@Lazarus-931 lgtm

Yavaren (approved):

lucasnewman:
@Lazarus-931 I'm sure this is better in the local maxima, but given all the churn here, I wonder if we should just have the standard model loader map all the filenames that would actually be used and then get the literal byte count, which should be 100% accurate?

Lazarus-931:
> @Lazarus-931 I'm sure this is better in the local maxima, but given all the churn here, I wonder if we should just have the standard model loader map all the filenames that would actually be used and then get the literal byte count, which should be 100% accurate?

yep, that works, closing this!

thx!

PR #557, opened 2026-09-15
keyboard shortcuts for tool callings in chat

Lazarus-931:
this is just small stuff to make the interface in chat easier.

this allows arrow buttons to work to pick image model

enter for selecting image model

escape to cancel the picker

escape to deny a tool approval, it only had enter to confirm before

only the picked model shows a "Use >" button now, the grey ones on every row were confusing

clicking anywhere on a downloaded row picks it

the pick stays put when the model list refreshes

first & second:

https://github.com/user-attachments/assets/32776749-7bb5-4a9c-a4fb-6f2b1bc971f6

third:

https://github.com/user-attachments/assets/5cbd781d-ce09-450c-84b2-17ddba7da445


lucasnewman (approved):
Nice!

PR #558, opened 2026-09-15
Microphone switch capture

Yavaren:
This PR replaces AVAudioEngine microphone capture with an input-only Core Audio unit, AUHAL. The goal was to improve recording startup, microphone switching, and cleanup, particularly when using Bluetooth headphones.

Microphone setup and teardown run outside the main actor. Recording only reports that it has started after audio actually arrives. The audio callback copies samples into a preallocated buffer, while file writing and level calculations happen outside that callback. Capture is rebuilt when the device or its format changes, with limited retries if startup or recovery fails. An unavailable selected microphone falls back to the system default.

Start and stop operations now wait for the previous capture connection to be released. Recording identifiers prevent callbacks from an older recording from changing a newer one. The existing file writer still uses the format of the delivered audio, handles format changes, and preserves captured audio when recovery fails. 

The audio page now has explicit Test microphone and Stop test buttons. Opening the page no longer activates the microphone automatically. Testing stops when recording begins or the user leaves the page, preventing the preview from keeping Bluetooth headphones in microphone mode between recordings.

All 37 regression tests passed. The capture tests also passed under Thread Sanitizer, and the 200-case reproduction of the deletion race returned 200 valid files with none deleted. Hardware testing covered repeated recordings, rapid cancellation, microphone preview, three headset speech recordings, and four input switches during one recording. Headphone output returned to 48 kHz roughly 2.1 seconds after the speech tests.


lucasnewman:
@Yavaren What's the purpose of this overall / what problem are you trying to solve? The HAL is much lower level and difficult to get right. I'm not sure this is the right direction.

Yavaren:
> @Yavaren What's the purpose of this overall / what problem are you trying to solve? The HAL is much lower level and difficult to get right. I'm not sure this is the right direction.

I've been trying to solve the problem with AV, but could not really find a good solution so found some working solutions online with AUHAL. There were other options with Objective-C bridges, but desided the current option would work best. 

The problem was the mic setup when starting recording with AirPods on. When you hit record for the first time, it would usually work just right, but then, it woud stall. See the video on Slack.

Yavaren:
> @Yavaren What's the purpose of this overall / what problem are you trying to solve? The HAL is much lower level and difficult to get right. I'm not sure this is the right direction.

Small edit*: the most popular robust solutions i found were either objective-c bridge or AUHAL

PR #559, opened 2026-09-15
Separate removed models in the dashboard model filter

lizen:
### Summary

The Dashboard's Model filter lists every model that has ever completed a request,
including models deleted long ago. With two models installed (Models → Installed),
the filter also showed two removed ones, mixed in with nothing marking them as
removed. Because `request_events` is never pruned, this list keeps growing as
models are downloaded, used, and deleted over time.

### Reproduction

1. Download a model, send a few chat requests, then delete it from Models → Installed.
2. Open Dashboard → the deleted model is still listed in the Model filter.

### Root cause

`DashboardViewModel.rebuildAvailableModels()` merged the local model scan with
`NativAnalyticsStore.fetchKnownModelIDs()`, which returns every model with a completed
request regardless of whether it is still installed or when it was last used.

### What this changes

- Groups the filter into **Installed** and **Previously used** sections.
- **Previously used** lists removed models that completed a request in the selected
  period, most recent first. The section title names the period
  (`Previously used · Last 7 days`), and a hint row such as
  `2 more used earlier — choose a longer period` counts the ones left out.
  All time lists every removed model.
- Changing the period keeps the selected model even when it has no activity in the new
  period. The model table says `This model has no requests in the selected period.`
  instead of the filter silently resetting to All.
- Adds a tooltip on the filter that explains the rule.
- Replaces `fetchKnownModelIDs()` with `fetchModelUsage()`, which also returns each
  model's last completed request (served by the existing `(model_id, completed_at)` index).

No analytics data is deleted; removed models stay reachable by choosing a longer period.

### Validation

- `make xcode-build` passes (XcodeGen 2.46.0; regenerating the project leaves no tracked changes).
- `DashboardModelFilterTests` (new, 2 tests), `TokenUsageModelPageTests` (9 tests), and
  `SoftwareUpdaterTests` (10 tests) pass with the same `xcodebuild ... test` invocation as Dev Build.
- Checked in a signed Debug build against real analytics data (2 installed models,
  2 removed models last used 22 and 40 days ago): Day hides both with `2 more used earlier`,
  Month lists one and hides one. All time lists both most recent first, and selecting a
  removed model under All time, then switching to Day keeps the selection.

### Screenshots
screenshot-period_all-time
screenshot-period_month
screenshot-period_week
screenshot-period_year
screenshot-period_day


lucasnewman (approved):
@lizen Nicely done, thank you!

PR #519, opened 2026-09-07
Model traces: record what each model was shown

Lazarus-931:
Records an append-only trace of every model call — system prompt by provenance, tools with their origin and schemas, thinking, tool calls and results — one trace per model instance per chat, readable from the chat page's existing side panel.

This is nice from a debugging standpoint, seeing what model as exposed to during each call and the subsequent info of each tool, mcp, skills is nice to see and help with understanding as well when model deviated, did something wrong, etc.

PR #560, opened 2026-09-15
1st of #519

Lazarus-931:
Extract shared SQLite access layer.


lucasnewman:
@Lazarus-931 This needs tests

PR #562, opened 2026-09-15
1st of #519

Lazarus-931:
Extract shared SQLite access layer.


Sources/Nativ/Features/Dashboard/NativAnalyticsStore.swift
```diff
@@ -773,188 +772,3 @@ final class NativAnalyticsStore {
...
-    case prepareFailed(String)
-    case stepFailed
-}
-
-private let sqliteTransientDestructor = unsafeBitCast(
-    -1,
-    to: sqlite3_destructor_type.self
-)
-
-private let schemaSQL = """
```
lucasnewman, line 894: How does the schema get set up with this change? Won't this break analytics?

lucasnewman (changes requested):
See comment.

Sources/Nativ/Features/Dashboard/NativAnalyticsStore.swift
Lazarus-931: didn't seem so during testing, the py server still makes and maintains the `Analytics.sqlite3` and its schema before writing any analytics, this only has read access as well!

lucasnewman:
@Lazarus-931 Looks like python creates the schema which is why it works. Codex review:

[P2] Keep prepare() uncached.
[SQLite.swift:106–109](https://github.com/Blaizzy/nativ/pull/562/files) returns the same mutable statement for identical SQL. Preparing another query can overwrite an existing caller’s bindings or reset its cursor. Cached aggregate statements also retain old read snapshots: I reproduced an analytics store returning 1 request after another connection committed 2, even across separate withReadSnapshot calls. The dashboard creates a fresh store per reload, which limits its exposure. Simplest fix: leave prepare() uncached and reserve caching for withStatement(), which resets statements on exit.

[P2] Update the existing dashboard test fixture.
The [switch to read-only initialization](https://github.com/Blaizzy/nativ/pull/562/files) breaks testDashboardLoadsEveryModelForTokenUsageWhileKeepingOtherMetricsGrouped: it relies on constructing the store to create its database and schema. The fixture should create those explicitly. Green CI misses this because [its Swift test command only selects SoftwareUpdaterTests](https://github.com/Blaizzy/nativ/blob/6d4038664f36cc419d2cd3bbad20c28ec7a0e8c0/.github/workflows/dev-build.yml#L46).

PR #572, opened 2026-09-16
Fix sleep recording

Yavaren:
Save interrupted recordings when the Mac or display sleeps, so stopping afterward no longer fails with “stream already stopped.”
Saved audio remains in Audio < Library for manual transcription, with a notice shown when the user returns. System sleep waits up to 30 seconds for saving to finish.


lucasnewman:
@Yavaren Codex review:

[P2] The interruption fallback can still delete captured audio — [SystemAudioMeetingRecorder.swift:157](https://github.com/Blaizzy/nativ/pull/572/files)
When prepareForInterruption() has run but stopCapture() throws another error, such as failedToStopAudioCapture, this branch calls audioWriter.cancel(). That already deletes the temporary file; skipping the subsequent removeItem cannot preserve it. My reproduction left neither temporary nor final audio. Marked interruptions should attempt to finalize the captured samples before cancellation.

[P2] An interruption during startup is silently dropped — [SystemAudioMeetingRecorder.swift:524](https://github.com/Blaizzy/nativ/pull/572/files)
self.stream is assigned only after await stream.startCapture(). If the delegate reports an interruption while that call is suspended, this guard discards it. Startup subsequently sets isRecording = true for the stopped stream. Track the pending stream and retain startup failures before awaiting capture startup.

Yavaren:
@lucasnewman fixes are out

lucasnewman (approved):
Looks good!

PR #573, opened 2026-09-16
Size models from the literal download byte count

Lazarus-931:
this is a much simpler solution than I proposed in #554

this does also fix the reported issue:

Same device(128gb)

## Before(main):

Screenshot 2026-09-16 at 8 51 40 AM

## After(this pr)
Screenshot 2026-09-16 at 8 49 59 AM


lucasnewman (approved):
Thanks!

Issue #570, opened 2026-09-16
Generated images do not appear in Artifacts until the app is reset


Blaizzy:
## Summary

A newly generated image does not appear on the **Artifacts** page during the current app session. The reporter says it becomes visible only after resetting the app, making successfully generated content appear missing from the gallery.

## Steps from the report

1. Generate an image in Nativ and wait for generation to finish.
2. Open the **Artifacts** page.
3. Observe that the new image is missing.
4. Reset the app and return to **Artifacts**; the image is now visible according to the report.

## Expected behavior

Successfully generated images become available in Artifacts as soon as generation completes, without needing to restart or reset the app.

## Actual behavior

The generated image stays absent from Artifacts until the app is reset.

## Acceptance criteria

- [ ] A completed image generation appears in Artifacts during the same app session.
- [ ] Artifacts updates both when the page is already open and when navigating to it after generation completes.
- [ ] Multiple completed generations appear without missing or duplicate entries.
- [ ] Generated images remain available after relaunching the app.

lucasnewman:
Resolved in https://github.com/Blaizzy/nativ/pull/574

PR #575, opened 2026-09-16
Unify uploaded and generated artifacts

Lazarus-931:
follow-up to #570 and #574, unifying uploaded files and generated images in artifacts, as well as:

making generated images marked properly when their origin is recorded, and showing “Unknown” for older files without reliable provenance for previous images generated

“Use in Chat” and image references reuse the same stored file and ID instead of creating another copy

preserving shared file identity and origin when importing chats

removing deleted artifacts from all linked chats and image sessions, with checks for active generation

#574 handles the live-refresh fix and is already merged; this PR contains the remaining unification and provenance changes


lucasnewman:
Resolved in the stack.

PR #578, opened 2026-09-16
2nd of #575

Lazarus-931:
Unify the artifact catalog, reuse stored files, and remove references across linked sessions.

Stacked on #577. Saves are sequential: failures can leave partial removal; remaining references can be retried.

Verified: local build and 35 tests passed


lucasnewman:
@Lazarus-931 Codex review:

[P2] Deleted artifacts can remain visible in another window.
The fingerprint is shared on disk, but known belongs to each window. After window A deletes an artifact and updates the fingerprint, window B can return its stale catalog because the underlying shared file still exists. The new mutationRevision only protects the window performing deletion. Remove this shortcut or associate each in-memory catalog with its own fingerprint. [ArtifactStore.swift:337–343](https://github.com/Blaizzy/nativ/pull/578/files)

[P2] Migration leaves obsolete copies of deleted files behind.
The rebuild switches to shared media references and removes the old cache-pruning code, but never cleans existing Caches/Nativ/Artifacts files. Those copies survive subsequent artifact deletion—even after shared-media garbage collection. This retains deleted content and wastes disk space. Add cleanup of obsolete copies during migration. [ArtifactStore.swift:346–350](https://github.com/Blaizzy/nativ/pull/578/files)

lucasnewman (approved):
LGTM

PR #587, opened 2026-09-18
Save local hardware diagnostics automatically with a 1 GB history limit

Blaizzy:
Nativ automatically saves local hardware diagnostic history for the public Mac model, chip, memory, CPU/GPU/ANE activity, disk health and sensors. Recording starts with the app, continues while the System tab is closed, and respects Pause and Resume. A fresh installation creates its history directory automatically without a marker or setup step.

Snapshots omit computer names, serial numbers, volume names, paths and raw sensor names. The recorder contains no networking.

The saved-history database is capped at **1 GB (1,000,000,000 bytes)** per device, with at most one snapshot per minute. History has no age or row-count cutoff. Oldest records are removed only when space is needed for another snapshot within the page limit. Free pages are reclaimed incrementally. SQLite's temporary rollback journal uses additional working space during writes and is removed after each transaction; writes pause below 1 GiB free disk space. A blocked reader cannot grow an unbounded WAL.

Validation:
- 20 selected XCTest cases passed in the existing `NativTests` target: six hardware-history tests, four observation-policy tests and ten software-update tests.
- Hardware coverage includes automatic creation, permissions, safe fields, missing values, restart throttling, pause, preserving old records and more than 10,080 rows below the budget, oldest-first eviction at the byte limit, low disk space and recovery after a blocked commit.
- The existing Dev Build workflow runs these tests in its current Xcode test step. No additional workflow or standalone test runner is needed.
- Unsigned Debug build/test passed with Xcode 27.0, reusing the existing packaged Python resource; no release packaging was performed.

Behavior and local validation commands are documented in `Docs/features/system-history.md`.


lucasnewman (approved):

lucasnewman:
@Blaizzy Needs verified commits to merge.

PR #588, opened 2026-09-18
Add structured request failure diagnostics and version context

Blaizzy:
Adds safe failure categories and app/runtime versions to local request records. Separates cancellations and correctly marks error events in HTTP 200 streams as failures. New diagnostic fields exclude request content and exception messages.

Validated with 49 script tests, a signed Debug build, and live success/failure checks. Tests run in the existing Dev Build workflow.


lucasnewman (approved):
LGTM

